Issue #19, opened 2021-10-06
Refactor and Unit Test ArchValidator


jondavid-black:
As a prototype developer I want to refactor my code and write unit tests so that my rapid prototype doesn't become painful technical debt in the near future.

jondavid-black:
I did quite a bit of refactoring on validator over the weekend.  It should be much easier to test now...but I didn't write tests.

PR #40, opened 2021-10-13
Add a few tests to validator

cvchaparro:
Relates to #19

I had more tests before the refactor but they are obsolete, now, so here are a few we can add. I still have to understand what all has changed in the validator before I can add more tests and continue refactoring it.


jondavid-black (approved):
the changes look good to me

PR #47, opened 2021-10-17
Update validator and tests

cvchaparro:
NOT READY FOR MERGE


jondavid-black:
It looks like a lot of the AaC structure is becoming hard coded into the validator.  How can we avoid that and use the AaC specification of itself to drive the validation?  The test would be to change the AaC spec and successfully validate itself without having to change anything in the validator.

PR #79, opened 2021-10-28
created a basic spec tree for testing

jondavid-black:
Not ready for review or merge yet.  Creating my first ever draft PR.  :)


jondavid-black:
The plugin and sample model are now in place.  Running validate fails because validate does not recognize the new spec extension on root.  This appears to be an issue with the validator, but I need to verify.  The test would be to apply an extension that adds a new field to the root type and then attempt to validate a YAML file using that root.

jondavid-black:
I continue to make some incremental but slow progress.  I believe I've fixed the bug that was blocking me (and manually tested that it is corrected).  Now I just need to find time to finish implementing the cross reference validation.

jondavid-black:
I've got the spec-validate working now against the calc model with manual tests.  I just need to get unit tests in place.  I find it a little challenging to unit test plugins because there's no return value for me to grab onto and I don't know how to capture things like console output in a unit test.

Coffee2Bits:
> 
> 
> I've got the spec-validate working now against the calc model with manual tests. I just need to get unit tests in place. I find it a little challenging to unit test plugins because there's no return value for me to grab onto and I don't know how to capture things like console output in a unit test.

Yeah, that is definitely a difficult part about the plugins. The pattern that I have been using is to break the logic of the plugin command out into smaller functions then test those "private" functions. If you had generated output to the filesystem for example then you could test using TempDirectory function from the `tempfile` package. ([example](https://github.com/jondavid-black/AaC/blob/6b5871167ea642c66439e94dee8f98a2be48f410/plugins/gen-design-doc/tests/test_aac_gen_design_doc_impl.py#L10))

Looking at the plugin command `spec_validate` I think you could change the sys.exit() below to raise an exception, then test your `spec_validate` command using [self.assertRaises()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) and expected failure/success cases.
https://github.com/jondavid-black/AaC/blob/1236437279411202dfd4232a2bd44e19099353b6/plugins/aac-spec/aac_spec_impl.py#L66-L70

plugins/aac-spec/aac-spec.yaml
```diff
@@ -1,18 +1,64 @@
+model:
+  name: aac-spec
+  description: aac-spec is a Architecture-as-Code plugin that enables requirement definition and trace in Arch-as-Code models.
```
cvchaparro, line 3:
```suggestion
  description: aac-spec is a Architecture-as-Code plugin that enables requirement definition and traceability in Architecture-as-Code models.
```

plugins/aac-spec/aac-spec.yaml
```diff
@@ -1,18 +1,64 @@
...
+      description: Validates spec traces within the AaC model
+      input:
+        - name: architecture_file
+          type: file
+          description:  The file to validate for spec cross-references.
+      acceptance:
+        - scenario: Valid spec traces are modeled.
+          given:
+            - The {{spec-validate.input.architecture_file}} contains a valid architecture specification.
+            - The {{spec-validate.input.parsed_model}} contains the parsed content from architecture_file.
```
cvchaparro, line 16: This `parsed_model` isn't in the inputs - but I thought `{{spec-validate.input.*}}` expected the `*` part (although it's not used, right now, obviously) to actually be an input in the model. If that's not the case, should we define how we're expecting this to be used?

plugins/aac-spec/aac-spec.yaml
```diff
@@ -1,18 +1,64 @@
...
+    add:
+      - name: requirements
+        type: RequirementReference[]
+---
 data:
   name: Specification
   fields:
     - name: name
       type: string
+    - name: abbrv
```
cvchaparro, line 51: Does `abbrv` stand for `abbreviation`? If so, I recommend we just say that instead for clarity-sake. If not, I recommend we use the full name (even more so) for clarity-sake.

plugins/aac-spec/aac-spec.yaml
cvchaparro: Looking further through the PR, I expect it was deleted so maybe we can just delete this line?

```suggestion
```

plugins/aac-spec/aac-spec.yaml
Coffee2Bits: I agree, I'll update this.

Coffee2Bits (approved):
Already reviewed with Cameron

plugins/aac-spec/aac_spec/aac_spec_impl.py
```diff
@@ -0,0 +1,90 @@
...
+
+plugin_version = "0.0.1"
+
+
+def spec_validate(architecture_file: str):
+    """
+    Validates spec traces within the AaC model.  If the model is invalid, print errors and exit with a code of 1.
+
+    Args:
+        architecture_file (file): The file to validate for spec cross-references.
```
Coffee2Bits, line 15:
```suggestion
        architecture_file (str): The file to validate for spec cross-references.
```

Coffee2Bits (changes requested):

Coffee2Bits (approved):

PR #94, opened 2021-11-05
Patch/create a new root 30

Coffee2Bits:
I've got the calculator.yaml validating at least. I wasn't too sure what more I could add or if what I did add was aligned with what you had in mind.

I made this PR to serve as a reference at least, if the changes aren't what you want feel free to not merge or just close the PR.

- Fixed mathmessagehandler.yaml file name
- Made some imports relative to overcome some parser issues (I'll play around later and ID if we need to make a bug issue)
- Added some missing behavior clauses and put in some behavior that seemed reasonable
- Updated the values in the aac-spec plugin to align with what's in the plugin yaml

### DIFF
--- a/model/calc/model/calculator.yaml
+++ b/model/calc/model/calculator.yaml
@@ -1,6 +1,6 @@
 import:
-  - mathmessagehandler.yaml
-  - mathlogger.yaml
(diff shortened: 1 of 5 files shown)


jondavid-black:
Thanks!

PR #97, opened 2021-11-05
Bug - Can't Validate Root Extension

### LINKED ISSUE #95: Cannot Validate when Extension is done to root
As a plugin developer I want to extend root so that I can create a new base structure in AaC YAML.

This is a bug currently blocking #30 .

The issue appears to be in the validator.  When getting the list of root types to support validation it only pulls core AaC root types and does not consider extensions.

Coffee2Bits:
Fixes the issue where the validator didn't recognize extensions to the root spec.
closes #95


jondavid-black (approved):
Thanks for fixing this so quickly.

PR #102, opened 2021-11-09
Simplify Model Behavior

Coffee2Bits:
- Removed the required behavior field.
- Added an external model for the User

### DIFF
--- a/model/flow/External.yaml
+++ b/model/flow/External.yaml
@@ -1,3 +1,3 @@
 model:
   name: User
-  behavior:
+  behavior: []
--- a/model/flow/External.yaml
+++ b/model/flow/External.yaml
@@ -1,0 +1,3 @@
+model:
+  name: User
+  behavior:
--- a/model/aac/AaC.yaml
+++ b/model/aac/AaC.yaml
@@ -142,7 +142,6 @@ data:
   required:
     - name
     - type
-    - acceptance
 
 ---
 data:
--- a/model/flow/flow.yaml
+++ b/model/flow/flow.yaml
@@ -1,6 +1,6 @@
 import:
   - ./System.yaml
-
+  - ./External.yaml
 usecase:
   name: Nominal flow of data through the system.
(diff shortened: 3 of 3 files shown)


model/flow/External.yaml
```diff
@@ -0,0 +1,3 @@
+model:
+  name: User
+  behavior:
```
Coffee2Bits, line 3: This is technically valid, but it looks weird to define an empty behavior for a model with no inputs/outputs.

cvchaparro: I think what we'd actually want to do is define it like this (i.e. an empty list of behaviors):

```yaml
model:
  name: User
  behavior: []
```

cvchaparro:
```suggestion
  behavior: []
```

cvchaparro:
Other than my previous comment, it looks good. My only question is regarding the validator tests. As they were, the tests all assumed behaviors required acceptance. This is no longer a valid assumption with this change so, I'd think we'd need to change those things.

For example:
- [Here](https://github.com/jondavid-black/AaC/blob/main/tests/test_validator.py#L83-L87) it might be good to add a behavior without an acceptance and make sure validation doesn't fail on that.

cvchaparro (changes requested):
Other than my previous comment, it looks good. My only question is regarding the validator tests. As they were, the tests all assumed behaviors required acceptance. This is no longer a valid assumption with this change so, I'd think we'd need to change those things.

For example:
- [Here](https://github.com/jondavid-black/AaC/blob/main/tests/test_validator.py#L83-L87) it might be good to add a behavior without an acceptance and make sure validation doesn't fail on that.

model/flow/External.yaml
jondavid-black: Why is it valid?  Behavior has required fields of name and type.  Given these fields aren't present shouldn't it be invalid?

cvchaparro (approved):

PR #90, opened 2021-11-11
I want Genplugin to generate a README file

cvchaparro:
As a plugin developer, I want Genplugin to generate an initial README so that I have a consistent outline and starting point for my plugin's README file.

AC:

- [x] Genplugin generates a README.md file
- [x] The README.md file is not overwritten if it already exists
- [x] The README should include as much information from the plugin's model as possible:
  - [x] Plugin Name
  - [x] Plugin commands + descriptions
  - [x] Plugin-defined definitions that aren't the plugin model (enum, data, extensions, etc)

Changes:

- Update `genplug.py` to not assume we're only generating python files.
- Add a `README.md` Jinja2 template that gets generated when a new plugin is created.

### DIFF
--- a/src/aac/templates/genplug/README.md.jinja2
+++ b/src/aac/templates/genplug/README.md.jinja2
@@ -2,16 +2,21 @@
 
 ## Plugin Commands
 
-### Shell Command
+{%- for command in commands %}
+
+### Command: {{command.name}}
+
+{{command.description}}
+
+Example shell usage:
 
 ```bash
-(.env) $ aac {{plugin.name}}
+(.env) $ aac {{command.name}} -h
 ```
 
-### Command arguments
+#### Command arguments
 
-- `-h, --help`: Shows a help message
-{%- for command in commands %}
+- `-h`, `--help`: Shows a help message
 {%- if command.input is defined %}
 {%- for command_input in command.input %}
 - `{{command_input.name}}`: {% if command_input.description is defined %}{{command_input.description}}{% else %}TODO: add description{%- endif %}
--- a/src/aac/templates/genplug/README.md.jinja2
+++ b/src/aac/templates/genplug/README.md.jinja2
@@ -1,0 +1,34 @@
+# {{plugin.name}}
+
+## Plugin Commands
+
+### Shell Command
+
+```bash
+(.env) $ aac {{plugin.name}}
+```
+
+### Command arguments
+
+- `-h, --help`: Shows a help message
+{%- for command in commands %}
+{%- if command.input is defined %}
+{%- for command_input in command.input %}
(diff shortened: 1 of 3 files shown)


src/aac/templates/genplug/README.md.jinja2
```diff
@@ -0,0 +1,34 @@
+# {{plugin.name}}
+
+## Plugin Commands
+
+### Shell Command
+
+```bash
+(.env) $ aac {{plugin.name}}
```
jondavid-black, line 8: Does this assume the plugin name is the same as the command name?  A plugin (ex: aac-puml) may have multiple commands, none of which match the plugin name.

src/aac/templates/genplug/README.md.jinja2
```diff
@@ -0,0 +1,34 @@
+# {{plugin.name}}
+
+## Plugin Commands
+
+### Shell Command
+
```
Coffee2Bits, line 6: Add some context here about the shell command with the plugin name -- that's how you'd invoke the plugin. For instance, add some info about how to run the plugin with the help command. `aac {{plugin.name}} -h`

src/aac/templates/genplug/README.md.jinja2
```diff
@@ -0,0 +1,34 @@
...
+### Command arguments
+
+- `-h, --help`: Shows a help message
+{%- for command in commands %}
+{%- if command.input is defined %}
+{%- for command_input in command.input %}
+- `{{command_input.name}}`: {% if command_input.description is defined %}{{command_input.description}}{% else %}TODO: add description{%- endif %}
+{%- endfor %}
+{%- endif %}
+{%- endfor %}
```
Coffee2Bits, line 20: It looks like the command scope isn't correct.

You should loop this entire section in your commands like:
```suggestion
{%- for command in commands %}
### Shell Command - {{command.name}}
{{command.description}}

Example shell usage:
\```bash  # Apparently I have to escape this or the suggestion breaks. 
(.env) $ aac {{command.name.name}} -h
\```

### Command arguments

- `-h, --help`: Shows a help message

{%- if command.input is defined %}
{%- for command_input in command.input %}
- `{{command_input.name}}`: {% if command_input.description is defined %}{{command_input.description}}{% else %}TODO: add description{%- endif %}
{%- endfor %}
{%- endif %}

{%- endfor %}
```

Coffee2Bits (changes requested):
Left a comment on the readme template.

src/aac/templates/genplug/README.md.jinja2
cvchaparro: Resolved with 1f38927

src/aac/templates/genplug/README.md.jinja2
cvchaparro: Resolved in 1f38927

Coffee2Bits (approved):

PR #107, opened 2021-11-15
Add Python types for GenPlugin Model Inputs

### LINKED ISSUE #76: Fix gen-plugin command generated implementation type hints and argument documentation
Currently, when generating a new plugin, the type hints provided for the function are incorrect. Specifically, the generated type hints correspond to Architecture-as-Code types rather than Python types. Additionally, the generated docstring does not make use of the provided description for `input` fields.

## Steps to Reproduce

1. Create a file from the following AaC model:
   ```yaml
   model:
     name: aac-test
     description: test description
     behavior:
       - name: test
         type: command
         description: test behavior description
         input:
           - name: some_inputs
             type: string[]
             description: description for some_inputs
           - name: another_input
             type: string
             description: description for another_input
         acceptance:
           - scenario: scenario
             when:
               - when
             then:
               - then
   ```
1. Generate a new plugin from the above model:
   ```bash
   (.env) $ aac gen-plugin test.yaml
   Do you want to generate an AaC plugin in the directory /Users/cameron/git/aac/? [y/n]y
   (.env) $ tree .
   .
   ├── __init__.py
   ├── aac_test.py
   ├── aac_test_impl.py
   ├── setup.py
   └── test.yaml

   0 directories, 5 files
   ```
1. View the contents of `aac_test_impl.py`:
   ```python
   """AaC Plugin implementation module for the aac-test plugin."""
   # This file is only initially generated by the aac gen-plugin, and it won't be overwritten if the file already exists.

   plugin_version = "0.0.1"


   def test(some_inputs: string[], another_input: string):
       """
       test behavior description.

       Args:
           some_inputs <string[]>: TODO add a helpful parameter description
           another_input <string>: TODO add a helpful parameter description
       """

       # TODO add implementation here
       raise NotImplementedError("test is not implemented.")
   ```
1. Observe type hints in the `def`'s parameter list and in the `Args:` section of the docstring. Additionally, the descriptions in the

## Expected Behavior

The generated plugin files reference Python types in parameter list type hints and docstring `Args:` sections. For example:

```python
"""AaC Plugin implementation module for the aac-test plugin."""
# This file is only initially generated by the aac gen-plugin, and it won't be overwritten if the file already exists.

plugin_version = "0.0.1"


# -> note that, here, we're actually using Python types instead of AaC primitives
def test(some_inputs: list[str], another_input: str):
    # -> note that, in the Args section the description from the YAML file is used and the referenced types for the input parameters are using Python types instead of AaC primitives.
    """
    test behavior description.

    Args:
        some_inputs <list[<str]>: description for some_inputs
        another_input <str>: description for another_input
    """

    # TODO add implementation here
    raise NotImplementedError("test is not implemented.")
```

### AC:
- [x] Implement a pythonic-type for primitive types

Coffee2Bits:
fixes #76 

New example output looks like this:
```python
"""AaC Plugin implementation module for the aac-gen-protobuf plugin."""
# This file is only initially generated by the aac gen-plugin, and it won't be overwritten if the file already exists.

plugin_version = "0.0.1"


def gen_protobuf(architecture_file: str, output_directory: str):
    """
    Generate protobuf messages from Arch-as-Code models.

    Args:
        architecture_file (str): The yaml file containing the data models to generate as Protobuf messages.
        output_directory (str): The directory to write the generated Protobuf messages to.
    """

    # TODO add implementation here
    raise NotImplementedError("gen_protobuf is not implemented.")

```

Note that in the above example that we're outputting with the pythonic string types.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -43,6 +43,7 @@ jobs:
           path: html_code_coverage/
       - name: Validate Core Spec and Plugin Specs
         run: |
+          source venv/bin/activate
           find . -name AaC.yaml -exec aac validate {} \;
           find plugins -name *.yaml -exec aac validate {} \;
       - name: Generate API Documentation
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -41,6 +41,10 @@ jobs:
         with:
           name: test-coverage-report
           path: html_code_coverage/
+      - name: Validate Core Spec and Plugin Specs
+        run: |
+          find . -name AaC.yaml -exec aac validate {} \;
+          find plugins -name *.yaml -exec aac validate {} \;
       - name: Generate API Documentation
         run: |
           source venv/bin/activate
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -44,7 +44,7 @@
             "program": "src/aac/__main__.py",
             "args": [
                 "validate",
-                "./plugins/aac-gen-protobuf/aac-gen-protobuf.yaml"
+                "./plugins/gen-protobuf/aac-gen-protobuf.yaml"
             ],
             "console": "integratedTerminal",
             "justMyCode": false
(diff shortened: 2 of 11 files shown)


jondavid-black (approved):
Looks good to me.

PR #110, opened 2021-11-18
Patch/jondavid black/create a new root 30

Coffee2Bits:
* Added tests to exercise all valid and invalid cases.
* Updated plugin structure
* Fixed plugin installation issue


plugins/gen-protobuf/aac_gen_protobuf/aac_gen_protobuf_impl.py
```diff
@@ -249,7 +249,3 @@ def _convert_camel_case_to_snake_case(camel_case_str: str) -> str:
     return snake_case_str
 
 
-class GenerateProtobufException(Exception):
```
cvchaparro, line 252: Was this change supposed to be here?

cvchaparro (commented):
Other than the one question, everything else looks good.

plugins/gen-protobuf/aac_gen_protobuf/aac_gen_protobuf_impl.py
Coffee2Bits: 🤦‍♂️

PR #91, opened 2021-11-18
I want Genplugin to generate my plugin with a proper directory structure

cvchaparro:
As a plugin developer, I want Genplugin to generate my plugin with a proper directory structure, so that my plugin's package is compatible with the package-based functions in Pluggy and Jinja2 .

AC:
- [x] Genplugin generates `setup.py` at the same level as the plugin AaC file
- [x] Genplugin generates a plugin package directory at the same level as the plugin AaC file
- [x] The plugin package directory has the same name as the plugin, and the directory name is snake_case (python style) (plugin name 'aac-gen-protobuf' would result in a directory named 'aac_gen_protobuf')
- [x] Genplugin generates the `plugin.py` file in the plugin package directory
- [x] Genplugin generates the `plugin_impl.py` file in the plugin package directory
- [x] Genplugin generates the `__init__.py` file in the plugin package directory

Changes:
- Update `genplug.py` to generate `plugin.py`, `plugin_impl.py`, and
  `__init__.py` into a package directory instead of the top-level plugin
  directory.

### DIFF
--- a/src/aac/genplug.py
+++ b/src/aac/genplug.py
@@ -165,6 +165,12 @@ def _compile_templates(parsed_models: dict[str, dict]) -> dict[str, list[Templat
 
     plugin_implementation_name = _convert_to_implementation_name(plugin_name)
 
+    template_parent_directories = template_parent_directories | {
+        "plugin.py.jinja2": plugin_name,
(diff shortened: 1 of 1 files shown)


Coffee2Bits (approved):
LGTM! 👍

PR #111, opened 2021-11-18
UI MVP

### LINKED ISSUE #7: User Interface MVP
As a system modeler I want a user interface that allows me to easily build a system model from top-down.

AC:
- [ ] Implement wireframes from #64 
- [x] Start the UI from an aac command `$ aac ui`

Coffee2Bits:
closes #7


src/aac/ui/menu_bar.py
```diff
@@ -0,0 +1,65 @@
...
+    Args:
+        root_window (Tk): the window to attach the menu to
+
+    Returns:
+        The menu bar.
+    """
+    menu_bar = Menu(root_window, tearoff=0)
+    _add_file_menu(menu_bar)
+    _add_edit_menu(menu_bar)
+    _add_view_menu(menu_bar)
```
cvchaparro, line 21: We spoke about removing the "View" menu but I think we could potentially use it for some/all of the following items (even if they're just disabled until they're implemented):

I know the first two are possible other places but I could see it being useful to be able to do it from the app menu. Either way we go with those, I think the last would be a good use for the view menu.

- We can use it as another place someone could switch from diagram to textual view - i.e. for power users that want to directly edit YAML.
- Change between "Diagram", "Properties", and "Imports" tabs.
- (Long-range but useful, IMHO) Change the types of diagram being displayed - i.e. app architecture, deployment architecture, network architecture, etc

.gitpod.Dockerfile
```diff
@@ -1,7 +1,8 @@
...
 
 ARG PYTHON_VERSION=3.9
 
 RUN sudo add-apt-repository ppa:deadsnakes/ppa -y
-RUN sudo apt install python${PYTHON_VERSION} -y 
-RUN sudo apt install python${PYTHON_VERSION}-venv -y
\ No newline at end of file
+RUN sudo apt install python${PYTHON_VERSION} -y
+RUN sudo apt install python${PYTHON_VERSION}-venv -y
+RUN sudo apt install python${PYTHON_VERSION}-tk -y
```
cvchaparro, line 8: Does `setup.py` need to be updated to include any extra dependencies?

src/aac/ui/menu_bar.py
```diff
@@ -0,0 +1,65 @@
...
+    _add_help_menu(menu_bar)
+
+    return menu_bar
+
+
+def _add_file_menu(root_menu: Menu) -> None:
+    """Configures the file menu dropdown."""
+    file_menu = Menu(root_menu, tearoff=0)
+
+    file_menu.add_command(label="New File", command=donothing)
```
cvchaparro, line 31: Might be good to call it "New Model", or something instead of "New File" or have a menu that comes off of "New..." that expands to give "Data", "Enum", "Model", or "Use Case" options.

src/aac/ui/menu_bar.py
```diff
@@ -0,0 +1,65 @@
+from tkinter import Menu, Tk
```
cvchaparro, line 1: General recommendation:

Disable items that aren't implemented - i.e. that `donothing` - so they can be clearly distinguished as "this doesn't work (yet)".

cvchaparro (commented):
Initial thoughts

.gitpod.Dockerfile
Coffee2Bits: Nope, Tkinter is apparently a python sys package and not a pypi one.

src/aac/cli.py
```diff
@@ -72,6 +72,11 @@ def _setup_arg_parser(
 
     # Built-in commands
     aac_commands = [
+        AacCommand(
+            "ui",
+            "Run the AaC Graphical User Interface",
+            ui.run_ui(),
```
Coffee2Bits, line 78:
```suggestion
            ui.run_ui,
```

src/aac/ui/menu_bar.py
Coffee2Bits: I think this should be moved to a discussion.

tox.ini
```diff
@@ -29,6 +29,7 @@ coverage-config = tox.ini
 [run]
 omit =
   *.jinja2
+  src/aac/ui/*
```
Coffee2Bits, line 32: I added this since we don't have an established method for testing the UI at the moment.

Coffee2Bits:
I was able to implement almost all of the wireframe, but there are some limitations within tkinter that need a lot more time and effort to implement. I wasn't able to implement scroll windows, it looks like it requires some extra tampering to implement ([example gist](https://gist.github.com/JackTheEngineer/81df334f3dcff09fd19e4169dd560c59)), and the buttons/marks aren't as stylized as the wireframe.

jondavid-black (approved):
I was able to manually test it using your instructions with GitPod.  Meets the AC for the MVP.  I'd like to hear about your experience and lessons learned building this out.  For example, is tkinter something you believe we can continue forward with?

src/aac/ui/menu_bar.py
cvchaparro: I'm fine with that.

cvchaparro (changes requested):
Other than the small change I suggested, I think it looks good. In the app itself, the formatting looks a bit strange (e.g. I think it would look better for the labeled input boxes to be aligned on the left and the right, rather than just on the right, etc) but those are things we can address later, too.

Coffee2Bits:
> Other than the small change I suggested, I think it looks good. In the app itself, the formatting looks a bit strange (e.g. I think it would look better for the labeled input boxes to be aligned on the left and the right, rather than just on the right, etc) but those are things we can address later, too.

I agree. That's been a lot of the struggle; Tkinter isn't nearly as easy to clean up as a web ui would be.

cvchaparro (approved):

PR #128, opened 2021-11-30
Remove the use of a virtual environment from the Github Actions CI/CD

### LINKED ISSUE #38: Remove the use of a virtual environment from the Github Actions CI/CD
Remove the use of a virtual environment from the Github Actions CI/CD. The pipeline currently leverages `venv` to create a virtual environment for dependencies and the application which isn't necessary for the pipeline since it's already a short-lived virtual environment. That being said, removing the use of venv causes issues with the pip module and installing the project's dependencies.

AC:
- [x] Github Actions don't leverage the venv Python module
- [x] The install dependencies Github Action step has pip installed
- [x] Github Actions can run tests, code checks, and build distributions without having the shell source a virtual environment

Coffee2Bits:
Closes #38 
- Removed the use of virtual environments
- Added necessary OS package installations
- Fixed up the validate smoke test to actually fail if the validate command fails
- Added a dependency analysis tool that I used to fix some dependency issues in the pipeline
- Cleaned up some of the dependency versions
- Added Python3.9 to the tox virtual test environment list


.github/workflows/python-app.yml
```diff
@@ -14,27 +14,34 @@ jobs:
...
       - name: Set up Python 3.9
         uses: actions/setup-python@v2
         with:
           python-version: 3.9
-      - name: Install dependencies
+      - name: Install Pip
+        uses: BSFishy/pip-action@v1
+        with:
+          packages: |
+            pip
```
Coffee2Bits, line 31:
```suggestion
          packages: pip
```

jondavid-black (approved):
Nice work!

PR #129, opened 2021-12-01
removed Abbrev requirement from YAML spec

BryanBailey11:


plugins/aac-spec/aac_spec/aac_spec.py
```diff
@@ -138,7 +135,6 @@ def get_base_model_extensions() -> str:
     type: string[]
   name: RequirementReference
   required:
```
jondavid-black, line 137: Also need to take out the abbreviation field.

Coffee2Bits (changes requested):
The changes look good. Can you make those changes in the `plugins/aac-spec/aac-spec.yaml` file as well? We haven't made the plugins rely on the yaml file as a source of truth yet so both the aac_spec.py and aac-spec.yaml need to be manually synced for now.

Coffee2Bits (approved):
Looks good. Meets the AC. 👍

cvchaparro:
LGTM - approved. Nice job, @BryanBailey11.

PR #130, opened 2021-12-02
Package AaC Core Spec DSL in the Distribution

### LINKED ISSUE #109: I want the AaC Core Spec DSL packaged in the executable.
As a user of AaC, I want AaC DSL files that describe distribution executables packaged in the executable, so that when I install the `aac` package it will contain the `AaC.yaml` file it needs in order to execute commands like `aac aac-core-spec`.

Example code for updating util.py to read in a package resource.
```python
    # get the AaC.yaml spec for architecture modeling
    with path(f"{__package__}.model", "AaC.yaml") as aac_model_file:
        AAC_MODEL = parser.parse_file(aac_model_file, False)
```

AC:
- [x] "AaC.yml" is moved to a directory module in the AaC project 
  - [x] The aac spec directory contains the AaC Core Spec file
- [x] The AaC setup.py is set to include the yaml file. 
  - [x] setup.py:ssetup() includes this line `package_data={'aac.model': ['AaC.yaml']}`
- [x] Update the core spec module to use `importlib.resources.path` to load the AaC.yaml file.
- [x] The pipeline exercises the distribution wheel
   - [x] `validate` works
   - [x] `aac-core-spec` works

Coffee2Bits:
Closes #109 

* Properly configured setup.py
  * Include data files (yaml, jinja)
  * Fixed packaging to work for editable and non-editable package installation
* Added a second pipeline job `test`
  * Installs the distribution wheel
  * Checks that the distribution works as expected
 * Removed bad dependencies that broke the gitpod env (Earlier mistake on my part)


.github/workflows/python-app.yml
```diff
@@ -63,8 +54,44 @@ jobs:
...
-      #   uses: pypa/gh-action-pypi-publish@master
-      #   with:
-      #     password: ${{ secrets.PYPI_API_TOKEN }}
+      - name: Upload Python Wheel
+        uses: actions/upload-artifact@v2
+        with:
+          name: aac_wheel
+          path: dist/
+
+  test:
```
cvchaparro, line 63: Could we call this smoke test, functional test, acceptance test, or something to make it more explicit what's going on. If we just call it test, then I'd think we should move (at least) the unit test stage to this job.

setup.py
```diff
@@ -59,16 +57,21 @@
 setup(
     name="aac",
     version="0.0.4",
+    description=(
+        "A distinctly different take on Model-Based System Engineering (MBSE) that allows a system modeller to define a system in simple yaml. "
+    ),
+    license="MIT",
```
cvchaparro, line 63: Did we for sure decide to license it with the MIT license, then?

cvchaparro (changes requested):

.github/workflows/python-app.yml
Coffee2Bits: Good point, it is nondescript

Coffee2Bits:
```suggestion
  Wheel Package Tests:
```

setup.py
Coffee2Bits: It's already the Jellyfish's license. I can remove the license for now.
https://github.com/NorthropGrumman/jellyfish/blob/master/LICENSE

Coffee2Bits:
```suggestion
    license="",
```

.github/workflows/python-app.yml
```diff
@@ -12,7 +12,6 @@ on:
 jobs:
   build:
```
Coffee2Bits, line 13:
```suggestion
  Build AaC Package and Docs:
```

.github/workflows/python-app.yml
```diff
@@ -10,9 +10,8 @@ on:
     branches: [main]
 
 jobs:
-  build:
+  Build AaC Package and Docs:
```
cvchaparro, line 13: Looks like GitHub doesn't like spaces in the name:

```suggestion
  Build-AaC-Package-and-Docs:
```

.github/workflows/python-app.yml
```diff
@@ -63,8 +54,44 @@ jobs:
...
-      #   uses: pypa/gh-action-pypi-publish@master
-      #   with:
-      #     password: ${{ secrets.PYPI_API_TOKEN }}
+      - name: Upload Python Wheel
+        uses: actions/upload-artifact@v2
+        with:
+          name: aac_wheel
+          path: dist/
+
+  Package Functional Tests:
```
cvchaparro, line 63: See above:

```suggestion
  Package-Functional-Tests:
```

.github/workflows/python-app.yml
```diff
@@ -63,8 +54,44 @@ jobs:
...
-      #     password: ${{ secrets.PYPI_API_TOKEN }}
+      - name: Upload Python Wheel
+        uses: actions/upload-artifact@v2
+        with:
+          name: aac_wheel
+          path: dist/
+
+  Package-Functional-Tests:
+    runs-on: ubuntu-latest
+    needs: build
```
cvchaparro, line 65:
```suggestion
    needs: Build-AaC-Package-and-Docs
```

cvchaparro (approved):

PR #134, opened 2021-12-02
Add extension plugins

### LINKED ISSUE #99: Allow Plugin with Extension Only
As a plugin developer I want to generate a plugin with no command so that I can extend the data model for my team.

AC:

- [x] Behavior must be optional when defining a model
- [x] Generate plugin must work when the model has no defined behavior
- [x] Installing the generated plugin must extend the data model as specified

cvchaparro:
Closes #99

Changes:

- Update `spec.yaml` to not require `behavior` when defining a model.
- Generate the plugin files inside a directory that uses the plugin name, except
  with underscores instead of hyphens.
- Do not require the `behavior` field when declaring models.
- Update the `plugin.py.jinja2` template to handle plugins with no commands.
- Update the `setup.py.jinja2` template to reference the plugin implementation,
  correctly.


README.md
```diff
@@ -225,6 +224,15 @@ $ aac --help
 # For example, to validate the spec.yaml file
 $ aac validate src/aac/spec/spec.yaml
 ```
+
+#### Running on MacOS
```
cvchaparro, line 228: Based on the changes to `setup.py` that @Coffee2Bits has in #130, we may not need this.

src/aac/spec/spec.yaml
```diff
@@ -178,19 +176,19 @@ data:
...
-    - name: source  # this should be a name of a participant (i.e. modelled system component)
+    - name: source
       type: string
-    - name: target  # this should be a name of a participant (i.e. modelled system component)
+    - name: target
       type: string
-    - name: action  # this as a behavior reference on the target or a subordinate use case definition
+    - name: action
       type: string
-    - name: if      # would create an alt in a sequence diagram
```
Coffee2Bits, line 189: I think these comments provide valuable context, why delete them?

src/aac/templates/genplug/plugin.py.jinja2
```diff
@@ -61,4 +64,3 @@ PLUGIN_EXTENSION_YAML = """
 {%- endfor %}
 """
 {%- endif %}
-
```
Coffee2Bits, line 64: Double check this, the generated python file might be missing its final blank line and might fail flake linting.

Coffee2Bits (changes requested):
Couple comments, but looks good otherwise! Good job! 👍

cvchaparro:
It's worth noting that I tried adding a functional-style test using pip as a
module but my attempts were in vain, and so I created #135 to address testing
the installation of a plugin and confirming we have access to the commands,
extensions, etc.

src/aac/spec/spec.yaml
cvchaparro: I don't really see what valuable context this provides, honestly. That said, if we chose to keep this for documenting the spec, we should probably add documentation to the entire spec. To that end, maybe we put the description key as an optional part of `Field` so that this kind of information can be provided and used in generation, etc.

src/aac/templates/genplug/plugin.py.jinja2
cvchaparro: I didn't mean to delete that newline. Good catch.

cvchaparro: Resolved in 91dc894

src/aac/spec/spec.yaml
Coffee2Bits: Well, I have already done that in GenPlugin so it's doable now. 

https://github.com/jondavid-black/AaC/blob/2445628738d69188d4f6aafdd0482500209ecf89/src/aac/genplug.py#L87-L94

cvchaparro: That's where I got the idea. 😉

Coffee2Bits (approved):

src/aac/spec/spec.yaml
cvchaparro: Resolved with #136

PR #143, opened 2021-12-07
140-move-plugins

### LINKED ISSUE #140: Move 1st party plugins moved into the AaC Repository
Move 1st party plugins moved into the AaC Repository.

AC:
- [x] Change the `plugins` python module from `plugins.py` to a directory module
  - [x] Move the functions in `plugins.py` to `__init__.py` in the `plugins` directory
- [x] Create a `plugins` submodule in the `tests` directory
- [x] For each plugin in the the plugins directory
   - [x]  Move the plugin's source into a directory module `aac.plugins.{plugin_name}`
   - [x] Move the plugin readme into `aac/docs/docs/plugins`
   - [x] Move the plugin tests into the `tests.plugins.{plugin_name}` directory module
   - [x] The AaC DSL describing the plugin is moved into the plugin package
   - [x] The plugin hooks that return the plugin's AaC yaml returns the value from the plugin yaml, not a hardcode string
   - [x] The plugin is registered in [plugins/__init__.py:get_plugin_manager()](https://github.com/jondavid-black/AaC/blob/2e36baa437857c4b3a591c70b9e2a997e8ef42ff/src/aac/plugins.py#L7-L22)
- [x] All 1st party plugins are updated to the new scheme
   - [x] aac-plantuml
   - [x] aac-spec
   - [x] gen-design-doc
   - [x] gen-gherkin-behaviors
   - [x] gen-protobuf
   - [x] genJson
   - [x] genPlugin
- [x] All 1st party plugin unit tests are run
- [x] Code coverage includes 1st party plugins

Coffee2Bits:
Closes #140.

* UPDATED AAC PACKAGE AND MOVED PLUGIN HOOKS TO THE AAC.PLUGINS PACKAGE TO AVOID CIRCULAR DEPENDENCIES
* Updated hookimpl names to remove the `aac` package name as the new FQN would have been `aac.plugins.hookimpl`
* Moved plugins into the python source directory
* UPDATED AAC-SPEC to SPECIFICATIONS to avoid confusion with the AaC Core Spec
* Updated plugin names for better consistency and removed redundant `aac` descriptors from names
* Removed 1st party plugin versions since they'll always be versioned by the AaC package
* Updated package setup to include all plugin model yaml files
* Updated all plugins to load their model definition from their AaC DSL file.
* Updated template engine loading to ignore non-jinja2 files in the templates directory.
* Updated pipeline smoke test with new plugin model file locations
* Added basic plugin models for plugins missing them
* Added some additional tests to raise the code coverage to 80% -- will need to update several plugins for testability to improve coverage amounts.


src/aac/plugins/plugin_manager.py
```diff
@@ -11,13 +8,30 @@ def get_plugin_manager() -> PluginManager:
     Returns:
         The plugin manager.
     """
+
+    from aac.plugins import hookspecs, PLUGIN_PROJECT_NAME
```
cvchaparro, line 12: I don't think it's generally a good idea to import within functions. Is there a specific reason we're doing that here, instead of at the top of the file?

src/aac/plugins/plugin_manager.py
```diff
@@ -29,6 +43,8 @@ def get_plugin_model_definitions():
     Returns:
         A list of plugin defined models.
     """
+    from aac import parser
```
cvchaparro, line 46: See previous comment about importing in a function definition.

src/aac/plugins/specifications/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+    spec_validate_arguments = [
+        AacCommandArgument("architecture_file", "The file to validate for spec cross-references."),
+    ]
+
+    plugin_commands = [
+        AacCommand(
+            "spec-validate",
+            "Validates spec traces within the AaC model.",
+            spec_validate,
+            spec_validate_arguments),
```
cvchaparro, line 31:
```suggestion
            spec_validate_arguments
        ),
```

src/aac/template_engine.py
```diff
@@ -18,9 +18,9 @@ def load_templates(package_name: str, template_directory: str = "templates") ->
     Returns:
         list of loaded templates
     """
-
```
cvchaparro, line 21: I think for inner functions we want to keep the space here.

tests/plugins/test_gen_gherkin_behaviors.py
```diff
@@ -177,8 +218,15 @@ def test_gen_gherkin_behaviors_with_gherkin_keyword_collisions(self):
...
 
 
 def _remove_first_word_in_string(string: str) -> str:
     return string.split(None, 1)[1]
+
+
+def _filter_command_behaviors(behavior_definition: dict):
+    if behavior_definition.get("type") == "command":
+        return behavior_definition
+    else:
```
cvchaparro, line 231: The else isn't needed:

```suggestion
```

tests/plugins/test_gen_gherkin_behaviors.py
```diff
@@ -177,8 +218,15 @@ def test_gen_gherkin_behaviors_with_gherkin_keyword_collisions(self):
...
+        return None
```
cvchaparro, line 232:
```suggestion
    return None
```

tests/plugins/test_gen_protobuf.py
```diff
@@ -0,0 +1,255 @@
...
+
+            # Assert each expected file is present
+            self.assertIn("data_a.proto", generated_file_names)
+            self.assertIn("data_b.proto", generated_file_names)
+            self.assertIn("data_c.proto", generated_file_names)
+            self.assertIn("message_metadata_data.proto", generated_file_names)
+            self.assertIn("message_type.proto", generated_file_names)
+
+            # Assert data_a.proto contents
+            with open(os.path.join(temp_dir, "data_a.proto")) as data_a_proto_file:
```
cvchaparro, line 30: I think we could extract a lot of the duplication here into a separate function. Maybe something like:

```python
def check_generated_file_contains(path: str, contents: list[str]) -> None:
    with open(path) as generated_file:
        generated_file_contents = generated_file.read()
        for string in contents:
            self.assertIn(string, generated_file_contents)
```

And then we could call it like this:

```python
check_generated_file_contains(os.path.join(temp_dir, "data_a.proto"), [
    "import \"message_metadata_data.proto\"",
    "import \"message_type.proto\"",
    "MessageMetadataData metadata",
    # etc
])
```

Especially since this is a common pattern we have for testing, we may want to make this available for other tests, as well.

src/aac/plugins/gen_gherkin_behaviors/__init__.py
```diff
@@ -31,3 +33,15 @@ def get_commands() -> list[AacCommand]:
...
+
+@hookimpl
+def get_base_model_extensions() -> str:
+    """
+    Returns the CommandBehaviorType modeling language extension to the plugin infrastructure.
+
+    Returns:
+        string representing yaml extensions and data definitions employed by the plugin
+    """
+    with resources.open_text(__package__, "gen_gherkin_behaviors.yaml") as plugin_model_file:
```
cvchaparro, line 46: Since this is a common pattern, we might do better extracting this into a helper function in `plugins/__init__.py` and just call it passing the name of the package and the name of the file - I think in that way it will be cleaner and less error prone.

cvchaparro (changes requested):
Nice work, overall it looks pretty good. There are a few things I think we could improve, here, though.

src/aac/plugins/gen_json/__init__.py
```diff
@@ -0,0 +1,43 @@
+"""Generated AaC Plugin hookimpls module for the aac-gen-protobuf plugin."""
```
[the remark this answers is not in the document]
Coffee2Bits: THanks for catching this! 😅

tests/plugins/test_gen_protobuf.py
Coffee2Bits: I agree. We should make some testing utilities.

I'll make a story for us to implement some testing utilities; be sure to think about some other functions that'd be helpful.

src/aac/plugins/gen_plant_uml/gen_plant_uml_impl.py
```diff
@@ -4,8 +4,6 @@
 from aac import parser, util
 from aac.validator import validation
 
-plugin_version = "0.0.2"
```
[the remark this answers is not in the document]
Coffee2Bits: "Removed 1st party plugin versions since they'll always be versioned by the AaC package"

1st party plugins don't need versions since their version is inherently linked to the AaC package.

src/aac/plugins/plugin_manager.py
Coffee2Bits: It was previously due to a circular import, but upon double checking it looks like I must have solved the core issue elsewhere.

Coffee2Bits: Nevermind, this issue is that there is a circular dependency loop validator -> plugins -> plugin with validator imported -> validator

```
ImportError: Failed to import test module: test_validator
Traceback (most recent call last):
  File "/home/runner/work/AaC/AaC/.tox/py39/lib/python3.9/site-packages/nose2/plugins/loader/discovery.py", line 201, in _find_tests_in_file
    module = util.module_from_name(module_name)
  File "/home/runner/work/AaC/AaC/.tox/py39/lib/python3.9/site-packages/nose2/util.py", line 77, in module_from_name
    __import__(name)
  File "/home/runner/work/AaC/AaC/tests/test_validator.py", line 5, in <module>
    from aac.validator import ValidationError, _validate, validation
  File "/home/runner/work/AaC/AaC/src/aac/validator.py", line 12, in <module>
    from aac import plugins, util
  File "/home/runner/work/AaC/AaC/src/aac/plugins/__init__.py", line 8, in <module>
    from aac.plugins import plugin_manager
  File "/home/runner/work/AaC/AaC/src/aac/plugins/plugin_manager.py", line 5, in <module>
    from aac.plugins import (
  File "/home/runner/work/AaC/AaC/src/aac/plugins/gen_json/__init__.py", line 9, in <module>
    from aac.plugins.gen_json.gen_json_impl import print_json
  File "/home/runner/work/AaC/AaC/src/aac/plugins/gen_json/gen_json_impl.py", line 5, in <module>
ImportError: cannot import name 'validation' from partially initialized module 'aac.validator' (most likely due to a circular import) (/home/runner/work/AaC/AaC/src/aac/validator.py)
```

src/aac/__init__.py
```diff
@@ -1,11 +0,0 @@
-from pluggy import HookimplMarker, HookspecMarker
```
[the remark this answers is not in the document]
Coffee2Bits: Plugin-related items were moved to `aac.plugins` so we don't need to expose it here in the `aac` package.

I think exporting `run_cli` was to combat circular dependencies, but it seems to have been resolved.

src/aac/template_engine.py
```diff
@@ -18,9 +18,9 @@ def load_templates(package_name: str, template_directory: str = "templates") ->
     Returns:
         list of loaded templates
     """
```
Coffee2Bits, line 20:
```suggestion
    """
    
```

src/aac/template_engine.py
Coffee2Bits: Done

tests/plugins/test_gen_protobuf.py
Coffee2Bits: The related issue is #158. The topic is a draft for now.

src/aac/plugins/gen_gherkin_behaviors/__init__.py
Coffee2Bits: I'd be more inclined to agree if this wasn't a two-line boilerplate pattern. 

There is also the consideration of injecting a circular dependency if a plugin relies on the `plugins` package, but the plugin is initialized as part of the `plugins` initialization where the plugins are registered.

cvchaparro: A two-line boilerplate pattern is still more than a one-line boilerplate pattern - i.e. a function call.

Regarding putting it into `plugins/__init__.py` causing circular import issues, that makes sense but couldn't that be resolved by putting it into another file in the `plugins/` directory? Maybe we can put it in `plugins/common.py`, or something similar to avoid circular imports?

src/aac/template_engine.py
```diff
@@ -18,9 +18,10 @@ def load_templates(package_name: str, template_directory: str = "templates") ->
...
-
+
```
cvchaparro, line 21:
```suggestion

```

src/aac/plugins/gen_gherkin_behaviors/__init__.py
Coffee2Bits: Ok, I've implemented a new module and function to thin the boilerplate a little.

cvchaparro (approved):
Looks good.

cvchaparro:
Can you fix the merge conflict with the pipeline?

PR #148, opened 2021-12-09
137 error on empty list for required item

### LINKED ISSUE #137: Update validator to require non-empty lists if the list is required
As a modeler, I want the validator to return a validation error if I provide an
empty list for a list-type field that is required so that I can be sure that my
model has all the information it needs.

## AC

- [ ] The validator is returns an error if a model with a list type field is `required` but there are no entries.

## Notes

- Related #104

cvchaparro:
Closes #137 

Changes:

- Generate error message if required list field is empty.

### DIFF
--- a/tests/test_validator.py
+++ b/tests/test_validator.py
@@ -150,28 +150,6 @@ class ValidatorTest(TestCase):
         for d in test_data:
             assert_model_is_invalid(self, d, pattern)
 
-    def test_valid_usecases_with_only_participants_pass_validation(self):
-        name = self.MODEL_NAME
-        fields = self.VALID_FIELDS
-
-        usecases = [
-            usecase(name=name, participants=fields[:i], steps=[]) for i in range(len(fields))
-        ]
-
-        for u in usecases:
-            assert_model_is_valid(self, u)
-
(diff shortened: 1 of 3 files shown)


tests/test_validator.py
```diff
@@ -145,41 +145,21 @@ def test_data_with_undefined_references_to_required_fields_fails_validation(self
...
         test_data = [
-            data(name=name, fields=fields[:i], required=[required[i]]) for i in range(len(fields))
+            data(name=name, fields=fields[:i], required=[required[i]])
+            for i in range(1, len(fields))
         ]
 
         for d in test_data:
             assert_model_is_invalid(self, d, pattern)
 
-    def test_valid_usecases_with_only_participants_pass_validation(self):
```
Coffee2Bits, line 154: Were these removed because the tests are invalid with empty required lists?

Coffee2Bits (commented):
Looks good, will approve once we resolve the question I posted.

tests/test_validator.py
cvchaparro: Yes

Coffee2Bits (approved):

Issue #147, opened 2021-12-09
Handle plugin execution results consistently


cvchaparro:
## Description

As a plugin developer, I want to be able to signal exception conditions that occur in my plugin without effecting the rest of the system so that errors can be propogated to the CLI/GUI and displayed to the user.

## AC

- [x] A class exists that represents a plugin execution result and it is available for use by all plugins.
- [x] GenPlugin is updated to have plugin commands return an instance of this class.
- [x] The CLI is updated to handle this returned execution result object.
  - [x] For success, print/log a success message.
  - [x] For failure, print/log a failure message and exit with non-zero status code.
- [x] At a minimum, the class has some way to provide the following information to whatever receives the result:
  - [x] `name`: the name of the plugin whose execution result is being handled.
  - [x] `status_code`: the status code to return from the CLI, for example.
  - [x] `errors`: a list of error messages that can be provided to the user so they know what went wrong.
- [x] The `status_code` is not just a number that users must know. That is, we don't want something like this:
    ```python
    result = command.callback(**kwargs)
    if result.status_code != 0:
        print('failure')
    ```
    Rather, do something like this:
    ```python
    result = command.callback(**kwargs)
    if result.execution_failed()
        print('failure')
    ```
    Or even like this:
    ```python
    result = command.callback(**kwargs)
    if result.status_code != PluginExecutionStatusCode.SUCCESS:
        print('failure')
    ```
- [x] If a plugin fails in the pipeline, the pipeline fails.

Coffee2Bits:
Added two sub-bullets for some extra clarity on what we're expecting.
```
* The class contains an error code status
* The class contains a list of errors
```

I think this is a pretty important issue for us to complete since it'll help us improve the testability of plugins as well as the benefits of running them in the UI that you pointed out. I've gone ahead and added it to sprint 6 -- but we can discuss it during planning.

cvchaparro:
Yeah, I think it is important. That said, I think those bullets are covered by AC item 4, no?

Coffee2Bits:
> Yeah, I think it is important. That said, I think those bullets are covered by AC item 4, no?

Yeah, I'm an idiot. I've removed them.

cvchaparro:
For the last AC, see: https://github.com/jondavid-black/AaC/runs/4722270887?check_suite_focus=true#step:10:1

PR #160, opened 2021-12-14
added "Validate Example Model" and "Test gen-plugin" jobs to pipeline

BryanBailey11:
Also updated main-branch.yml and pull-request.yml with new Shaw number.

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -35,9 +35,9 @@ jobs:
 
     - name: Test gen-plugin
       run: |
-        mkdir -p AaC/model/test
-        cp AaC/plugins/aac-spec/aac-spec.yaml AaC/model/test
-        aac gen-plugin AaC/model/test/aac-spec.yaml
+        mkdir gen_plugin_generation_test
+        cp  src/aac/spec/spec.yaml gen_plugin_generation_test
+        aac gen-plugin gen_plugin_generation_test/spec.yaml
 
     - name: Print out Core Spec
       run: aac aac-core-spec
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -37,7 +37,7 @@ jobs:
       run: |
         mkdir gen_plugin_generation_test
         cp  src/aac/plugins/gen_protobuf/gen_protobuf.yaml gen_plugin_generation_test
-        (yes y ||:) | aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml
+        echo y | aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml
 
     - name: Print out Core Spec
       run: aac aac-core-spec
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -37,7 +37,7 @@ jobs:
       run: |
         mkdir gen_plugin_generation_test
         cp  src/aac/plugins/gen_protobuf/gen_protobuf.yaml gen_plugin_generation_test
-        yes y | aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml
+        yes y>/dev/null | aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml
 
     - name: Print out Core Spec
       run: aac aac-core-spec
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -37,7 +37,10 @@ jobs:
       run: |
         mkdir gen_plugin_generation_test
         cp  src/aac/plugins/gen_protobuf/gen_protobuf.yaml gen_plugin_generation_test
-        aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml <<< "y"
+        spawn aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml
+        expect "Do you want to generate an AaC plugin in the directory /home/runner/work/AaC/AaC/gen_plugin_generation_test? [y/n]"
+        send "y"
+        expect eof
 
     - name: Print out Core Spec
       run: aac aac-core-spec
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -29,5 +29,15 @@ jobs:
         find src -name spec.yaml -print0 | xargs -0 -n1 aac validate
         find src/aac/plugins -name *.yaml -print0 | xargs -0 -n1 aac validate
 
+    - name: Validate Example Model
+      run: |
+        find model/flow -name system.yaml -print0 | xargs -0 -n1 aac validate
+
+    - name: Test gen-plugin
+      run: |
+        mkdir -p AaC/model/test
+        cp AaC/plugins/aac-spec/aac-spec.yaml AaC/model/test
+        Aac gen-plugin AaC/model/test/aac-spec.yaml
+
     - name: Print out Core Spec
       run: aac aac-core-spec
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -31,7 +31,7 @@ jobs:
 
     - name: Validate Example Model
       run: |
-        find model/flow -name system.yaml -print0 | xargs -0 -n1 aac validate
+        aac validate model/flow/System.yaml
 
     - name: Test gen-plugin
       run: |
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -37,7 +37,7 @@ jobs:
       run: |
         mkdir gen_plugin_generation_test
         cp  src/aac/plugins/gen_protobuf/gen_protobuf.yaml gen_plugin_generation_test
-        printf 'y' | aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml
+        printf 'y\n' | aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml
 
     - name: Print out Core Spec
       run: aac aac-core-spec
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -36,7 +36,7 @@ jobs:
     - name: Test gen-plugin
       run: |
         mkdir gen_plugin_generation_test
-        cp  src/aac/plugins/gen_protobuf.yaml gen_plugin_generation_test
+        cp  src/aac/plugins/gen_protobuf/gen_protobuf.yaml gen_plugin_generation_test
         yes y | aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml
 
     - name: Print out Core Spec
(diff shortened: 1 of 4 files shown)


.github/workflows/main-branch.yml
```diff
@@ -11,7 +11,7 @@ jobs:
 
   Run_Functional_Tests:
     needs: Build_Package_and_Artifacts
-    uses: jondavid-black/AaC/.github/workflows/functional-tests.yml@9e0ff39886f7a0ce54e6eac6c6f2d78f9f9f1254
+    uses: jondavid-black/AaC/.github/workflows/functional-tests.yml@eb8a6a625e759ffc837569ac270577096e8a31cf
```
Coffee2Bits, line 14: Try this:
```suggestion
    uses: jondavid-black/AaC/.github/workflows/functional-tests.yml@29dcbdd69f2a15d77ade3f30a6f9779f86adf627
```

.github/workflows/pull-requests.yml
```diff
@@ -10,4 +10,4 @@ jobs:
 
   Run_Functional_Tests:
     needs: Build_Package_and_Artifacts
-    uses: jondavid-black/AaC/.github/workflows/functional-tests.yml@9e0ff39886f7a0ce54e6eac6c6f2d78f9f9f1254
\ No newline at end of file
+    uses: jondavid-black/AaC/.github/workflows/functional-tests.yml@eb8a6a625e759ffc837569ac270577096e8a31cf
```
Coffee2Bits, line 13: Try this:
```suggestion
    uses: jondavid-black/AaC/.github/workflows/functional-tests.yml@29dcbdd69f2a15d77ade3f30a6f9779f86adf627
```

.github/workflows/functional-tests.yml
```diff
@@ -29,5 +29,15 @@ jobs:
         find src -name spec.yaml -print0 | xargs -0 -n1 aac validate
         find src/aac/plugins -name *.yaml -print0 | xargs -0 -n1 aac validate
 
+    - name: Validate Example Model
+      run: |
+        find model/flow -name system.yaml -print0 | xargs -0 -n1 aac validate
```
cvchaparro, line 34: We don't actually need  to use `find` for this since we're targeting one specific file:

```suggestion
        aac validate model/flow/system.yaml
```

.github/workflows/functional-tests.yml
```diff
@@ -29,5 +29,15 @@ jobs:
...
+
+    - name: Test gen-plugin
+      run: |
+        mkdir -p AaC/model/test
+        cp AaC/plugins/aac-spec/aac-spec.yaml AaC/model/test
+        Aac gen-plugin AaC/model/test/aac-spec.yaml
```
cvchaparro, line 40:
```suggestion
        aac gen-plugin AaC/model/test/aac-spec.yaml
```

cvchaparro (changes requested):

.github/workflows/functional-tests.yml
BryanBailey11: okay, I see the difference. I was leveraging working code. Thanks for the suggestion.

.github/workflows/functional-tests.yml
BryanBailey11: Understood, will remember for next time.

BryanBailey11:
it looks like we have changed the path and name for the aac-spec.yaml file. I will have to update the path. Would this be better to not hard code the path to this file?

.github/workflows/functional-tests.yml
```diff
@@ -29,5 +29,15 @@ jobs:
...
 
+    - name: Validate Example Model
+      run: |
+        aac validate model/flow/System.yaml
+
+    - name: Test gen-plugin
+      run: |
+        mkdir -p AaC/model/test
+        cp AaC/plugins/aac-spec/aac-spec.yaml AaC/model/test
+        aac gen-plugin AaC/model/test/aac-spec.yaml
```
Coffee2Bits, line 40: This should address the pipeline issue. I also suggest changing the temporary directory name to be verbose rather than adding complexity to the directory structure via multiple levels.
```suggestion
        mkdir gen_plugin_generation_test
        cp  src/aac/spec/spec.yaml gen_plugin_generation_test
        aac gen-plugin gen_plugin_generation_test/spec.yaml
```

BryanBailey11: Thanks again

Coffee2Bits:
Looks like you'll need to update the reference sha again.

.github/workflows/functional-tests.yml
```diff
@@ -29,5 +29,15 @@ jobs:
...
 
+    - name: Validate Example Model
+      run: |
+        aac validate model/flow/System.yaml
+
+    - name: Test gen-plugin
+      run: |
+        mkdir gen_plugin_generation_test
+        cp  src/aac/spec/spec.yaml gen_plugin_generation_test
+        aac gen-plugin gen_plugin_generation_test/spec.yaml
```
Coffee2Bits, line 40: This cli tool, `yes` works on my local system to get past the user input.
```suggestion
        yes y | aac gen-plugin gen_plugin_generation_test/spec.yaml
```

.github/workflows/functional-tests.yml
```diff
@@ -29,5 +29,15 @@ jobs:
...
 
+    - name: Validate Example Model
+      run: |
+        aac validate model/flow/System.yaml
+
+    - name: Test gen-plugin
+      run: |
+        mkdir gen_plugin_generation_test
+        cp  src/aac/spec/spec.yaml gen_plugin_generation_test
+        yes y | aac gen-plugin gen_plugin_generation_test/spec.yaml
```
cvchaparro, line 40: This produced a "gen-plugin" error: https://github.com/jondavid-black/AaC/runs/4540414112?check_suite_focus=true#step:8:9

We but the pipeline passed - we should make sure we exit with a failure (so that the pipeline fails) when this command fails or it could cause us problems later on. That said, I think we may be able to tackle this correctly with #147.

cvchaparro: I've added a new AC to that story to make sure we check that.

Coffee2Bits: We could make an issue to create a test bash script that runs the plugin and then asserts the expected files were generated.

This leads into another thought; I've had several QOL improvement thoughts that could be implemented with simple scripts, but I've been reluctant to introduce a scripts directory. I'm a lot more amicable to supporting a scripts directory if it includes functional test scripts like what this step needs.

cvchaparro: I'm not opposed to having/supporting a scripts directory to give us QOL improvements for developing the core. If it makes maintenance/testing easier and/or more effective, I think it could be worth it assuming they're small scripts. If the scripts in this directory get big enough that we start needing tests for them, etc then that could start to outweigh the benefit, though, so we'll have to keep an eye on that for sure.

That said, I think what I'm doing with my next story will handle this specific case of making sure the pipeline fails if any error is generated. It obviously won't cover checking files, etc but we could create the story for that, still.

Coffee2Bits: Changes to make:
1. Use the [gen protobuf plugin ](https://github.com/jondavid-black/AaC/blob/main/src/aac/plugins/gen_protobuf/gen_protobuf.yaml) yaml instead 
2. If broken pipe persists, this may be a [solution](https://stackoverflow.com/questions/20573282/hudson-yes-standard-output-broken-pipe)

.github/workflows/functional-tests.yml
```diff
@@ -29,5 +29,15 @@ jobs:
...
 
+    - name: Validate Example Model
+      run: |
+        aac validate model/flow/System.yaml
+
+    - name: Test gen-plugin
+      run: |
+        mkdir gen_plugin_generation_test
+        cp  src/aac/plugins/gen_protobuf/gen_protobuf.yaml gen_plugin_generation_test
+        (yes y ||:) | aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml
```
Coffee2Bits, line 40:
```suggestion
        echo y | aac gen-plugin gen_plugin_generation_test/gen_protobuf.yaml
```

cvchaparro:
I don't think this is actually related to the `echo y | ...`, etc.  Rather, it looks like it's having some issues with encoding when [loading the templates](https://github.com/jondavid-black/AaC/runs/4553811711?check_suite_focus=true#step:8:25). We may need to make sure the encoding is 'utf-8' for all our files.

cvchaparro:
> I don't think this is actually related to the `echo y | ...`, etc. Rather, it looks like it's having some issues with encoding when [loading the templates](https://github.com/jondavid-black/AaC/runs/4553811711?check_suite_focus=true#step:8:25). We may need to make sure the encoding is 'utf-8' for all our files.

Hm, looks like the files in `src/aac/templates/` are all `us-ascii` encoded. I would assume this means it can be converted to `utf-8` without issue, though.

```console
√ aac % fd . 'src/aac/templates' | xargs -n1 file -I
src/aac/templates/__init__.py: text/plain; charset=us-ascii
src/aac/templates/gen_design_doc: inode/directory; charset=binary
src/aac/templates/gen_design_doc/__init__.py: inode/x-empty; charset=binary
src/aac/templates/gen_design_doc/datum.md.jinja2: text/plain; charset=us-ascii
src/aac/templates/gen_design_doc/macros.jinja2: text/plain; charset=us-ascii
src/aac/templates/gen_design_doc/model.md.jinja2: text/plain; charset=us-ascii
src/aac/templates/gen_design_doc/system-design-doc.md.jinja2: text/plain; charset=us-ascii
src/aac/templates/gen_design_doc/usecase.md.jinja2: text/plain; charset=us-ascii
src/aac/templates/gen_gherkin_behaviors: inode/directory; charset=binary
src/aac/templates/gen_gherkin_behaviors/__init__.py: inode/x-empty; charset=binary
src/aac/templates/gen_gherkin_behaviors/behavior.feature.jinja2: text/plain; charset=us-ascii
src/aac/templates/gen_protobuf: inode/directory; charset=binary
src/aac/templates/gen_protobuf/__init__.py: inode/x-empty; charset=binary
src/aac/templates/gen_protobuf/message.proto.jinja2: text/plain; charset=us-ascii
src/aac/templates/genplug: inode/directory; charset=binary
src/aac/templates/genplug/README.md.jinja2: text/plain; charset=us-ascii
src/aac/templates/genplug/__init__.py: text/plain; charset=us-ascii
src/aac/templates/genplug/__init__.py.jinja2: text/plain; charset=us-ascii
src/aac/templates/genplug/plugin.py.jinja2: text/plain; charset=us-ascii
src/aac/templates/genplug/plugin_impl.py.jinja2: text/plain; charset=us-ascii
src/aac/templates/genplug/setup.py.jinja2: text/plain; charset=us-ascii
src/aac/templates/genplug/test_plugin_impl.py.jinja2: text/x-script.python; charset=us-ascii
```

Coffee2Bits (approved):
Looks good. Good work!

cvchaparro (approved):

PR #161, opened 2021-12-16
136 make spec self documenting

### LINKED ISSUE #136: Make the spec self-documenting
As a system modeler, I want the spec to be self-documenting so that I can see
all the information related to a field in the field itself.

## AC

- [x] The spec is updated so that data fields have optional descriptions. (See
      `genplug.py` for the example descript field information.)
- [x] A clear description is added to all spec fields describing it's purpose
      and, possibly, offering an example if one would be useful.
- [x] The descript field is removed from the `genplug.py` plugin definition.

## Notes

- The `Step` data model had comments that were removed. Document these items so
  that a modeler can read a description of what they do, what they may be used
  for, etc.

  For reference, here are the original comments associated with the `Step` model:

  https://github.com/jondavid-black/AaC/blob/7567653eb1dc0e30170d5532024f1b3c1f31c694/src/aac/spec/spec.yaml#L177-L194
- Another benefit would be that we could generate documentation from the spec
  using the `description` fields to document each item.

cvchaparro:
Closes #136

Changes:

- Add the `description` property to the `data` definition.
- Remove the `DescriptField` extension from `gen_plugin.yaml`.
- Add descriptions for all items in the core spec.

### DIFF
--- a/src/aac/spec/spec.yaml
+++ b/src/aac/spec/spec.yaml
@@ -105,6 +105,8 @@ data:
       type: string
     - name: type
       type: string
+    - name: description
+      type: string
   required:
     - name
     - type
--- a/src/aac/spec/spec.yaml
+++ b/src/aac/spec/spec.yaml
@@ -179,10 +179,21 @@ data:
   fields:
     - name: name
       type: string
+      description: |
+        The name of the field.
+
+        Used to refer to the field, in certain cases.
     - name: type
       type: string
+      description: |
+        The type of the field.
+
+        Used to constrain the field's values to make it simpler to reason about.
     - name: description
       type: string
+      description: |
+        A description for the field to let users know what it's for and any
+        other useful information.
   required:
     - name
     - type
--- a/src/aac/spec/spec.yaml
+++ b/src/aac/spec/spec.yaml
@@ -350,8 +350,12 @@ data:
   fields:
     - name: condition
       type: string
+      description: |
+        A condition to test whether a sequence of steps should be performed.
     - name: steps
       type: Step[]
+      description: |
+        A list of steps to perform if the condition for the branch is "true".
   required:
     - condition
     - steps
--- a/src/aac/spec/spec.yaml
+++ b/src/aac/spec/spec.yaml
@@ -113,8 +113,13 @@ data:
   fields:
     - name: add
       type: Field[]
+      description: |
+        The list of fields to add to a `data' extension.
     - name: required
       type: string[]
+      description: |
+        The list of names of all the additional fields that are required for the
+        extended definition.
   required:
     - add
 ---
--- a/src/aac/spec/spec.yaml
+++ b/src/aac/spec/spec.yaml
@@ -7,8 +7,9 @@ data:
     - name: import
       type: string[]
       description: |
-        A list references to dependent models whose definitions should be
-        available for use by the current model.
+        A list of references to other models composing the current model. 
+        
+        Component model definitions must be visible from the current model's scope.
     - name: enum
       type: enum
       description: |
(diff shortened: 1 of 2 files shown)


src/aac/spec/spec.yaml
```diff
@@ -6,24 +6,60 @@ data:
   fields:
     - name: import
       type: string[]
+      description: |
+        A list references to dependent models whose definitions should be
+        available for use by the current model.
```
Coffee2Bits, line 11:
```suggestion
        A list of references to other models composing the current model. 
        
        Component model definitions must be visible from the current model's scope.
```

src/aac/spec/spec.yaml
```diff
@@ -6,24 +6,60 @@ data:
...
+        An example of when to use an `enum' is when you want to provide several
+        different options for a single value.
     - name: data
       type: data
+      description: |
+        A definition that represents a data value.
+
+        An example of when to use a `data' definition is when you want to define
+        some part of the system that can be an input and/or output to another
+        part of the system.
```
Coffee2Bits, line 26: I don't think this is quite accurate as `data` is _the_ keystone to our DSL and represents a general structure that acts as the backbone for the rest of the DSL. All of that to say that data is probably not the correct name for this key; we should have a discussion around that. 

For now, I'd make this more nebulous or maybe rather than focus on data as information, focus on data as shorthand for datastructure.

src/aac/spec/spec.yaml
```diff
@@ -56,8 +113,13 @@ data:
   fields:
     - name: add
       type: Field[]
+      description: |
+        The list of fields to add to a `data' extension.
```
Coffee2Bits, line 117:
```suggestion
        The list of fields to add to a `data` extension.
```

src/aac/spec/spec.yaml
```diff
@@ -103,8 +179,21 @@ data:
...
+      description: |
+        The name of the field.
+
+        Used to refer to the field, in certain cases.
     - name: type
       type: string
+      description: |
+        The type of the field.
+
+        Used to constrain the field's values to make it simpler to reason about.
```
Coffee2Bits, line 191:
```suggestion
        The type of the field.

        Used to constrain the field's values to make it simpler to reason about. Must 
        reference an existing definition by name.
```

src/aac/spec/spec.yaml
```diff
@@ -148,12 +263,21 @@ data:
   fields:
     - name: name
       type: string
+      description: |
+        The name of the system model.
```
Coffee2Bits, line 267:
```suggestion
        The name of the model.
```

src/aac/spec/spec.yaml
```diff
@@ -148,12 +263,21 @@ data:
...
     - name: description
       type: string
+      description: |
+        An explanatory description for the model including what the
+        component/system is modeleing and any other relevant information.
     - name: components
       type: Field[]
+      description: |
+        A list of the components that are part of the system.
```
Coffee2Bits, line 276:
```suggestion
        A list of models that are components of the system.
```

src/aac/spec/spec.yaml
```diff
@@ -148,12 +263,21 @@ data:
...
     - name: behavior
       type: Behavior[]
+      description: |
+        A list of behaviors that the system will perform.
```
Coffee2Bits, line 280:
```suggestion
        A list of behaviors that the model will perform.
```

Coffee2Bits (changes requested):
Some minor changes, but I'm surprised how much I love having those descriptions in the DSL. I think it provides a lot of context and will be beneficial.

Great work!

src/aac/spec/spec.yaml
cvchaparro: What do you think about changing it to this, instead:

```suggestion
        A list of behaviors that the system being modeled will perform.
```

Coffee2Bits: I like it!

Coffee2Bits (approved):

PR #165, opened 2021-12-16
AaC Spec Validation Update

### LINKED ISSUE #127: Update AaC Spec Validation to Use a Single Requirement Identifier
As an aac-spec plugin user, I want the AaC Spec Validation command updated to use single-string requirement identifiers, so that I can validate AaC Spec compliant models.

Per #113 the AaC Spec was updated with more succinct requirement IDs, but the story AC overlooked updating the AaC Spec plugin's validation and linking.

AC:
- [ ] AaC Spec Validation correctly recognizes and references the new requirement identifiers
- [ ] The yaml value returned by the aac_spec.py should be updated to match the plugin's yaml file
- [ ] The Calc model is updated with the new requirement identifiers
- [ ] The Calc model is successfully validated by the AaC Spec validation command

### LINKED ISSUE #153: Refactor the _run_spec_validation function to reduce complexity
## Description

As a maintainer, I want `_run_spec_validation` refactored to reduce complexity so that I can understand how to address bugs or add new functionality to the aac-spec plugin.

## AC

- [ ] The `# noqa: C901` linter comment is removed from the `aac_spec_impl.py` file.
- [ ] The `_run_spec_validation` function is broken into smaller functions.

## Notes

- The real test, here, will be whether the pipeline passes after removing the `noqa` comment.

Coffee2Bits:
Closes #127
Closes #153


tests/plugins/test_specifications_impl.py
```diff
@@ -0,0 +1,111 @@
...
+
+        with NamedTemporaryFile("w") as temp_spec:
+            temp_spec.write(INVALID_SPEC_MISSING_REQ_ID)
+            temp_spec.seek(0)
+
+            with self.assertRaises(AacSpecValidationException) as context_manager:
+                spec_validate(temp_spec.name)
+
+            validation_exception = context_manager.exception
+            self.assertIn("Invalid requirement id \"SUB-3\"", validation_exception.message)
```
cvchaparro, line 30: I don't think `context_manager` wouldn't be active in this scope, anymore, would it? Should we have this, instead?

```suggestion
                validation_exception = context_manager.exception
                self.assertIn("Invalid requirement id \"SUB-3\"", validation_exception.message)
```

cvchaparro (changes requested):
Overall looks good, there were just a couple things I had questions on. One final thing, though, would be to please merge `main` so you can remove the `# noqa: C901` comment from the `_run_spec_validation` function definition line since that's telling `flake8` to ignore the complexity warning.

tests/plugins/test_specifications_impl.py
Coffee2Bits: It is accessible outside the scope. Moving this in would hide the exception from the assert if it doesn't receive and exception.

cvchaparro: Well, I just learned something about context managers, then. Thanks!

cvchaparro (approved):

PR #169, opened 2021-12-27
Add PyPI Package Deployment 

### LINKED ISSUE #114: I want the AaC package deployed to pypi
As an AaC user, I want the packaged deployed to pypi, so that I can install it in my environments via pip.

NOTE: This issue should be completed after #109 so that any package deployed to pypi has the core AaC spec packaged correctly.

AC:

- [x] The deploy job is moved from the main-branch.yaml file to a manual workflow job
- [x] The deploy job is only trigger manually
- [x] The deploy job publishes the aac package to pypi.
- [ ] The aac package can be installed via `pip install aac`

Coffee2Bits:
Closes #114 

Due to the nature of how Gtihub Actions work, I forked the AaC repo in order to try and "test" the workflow before this PR. I attempted to use the test PyPI index `https://test.pypi.org/`, but the AaC package is already there and tied to JD's account. So, I've included the reference file and pipeline to demonstrate that the pipeline works up until it hits a user permissions error, which we shouldn't see here.

The mostly-successful GHA run be found here: 
https://github.com/Coffee2Bits/AaC/actions/runs/1613374176

The forked workflow file can be found here: https://github.com/Coffee2Bits/AaC/blob/37a1de534585f0f079abbd119d13233b31d30043/.github/workflows/deploy-artifacts.yml

Changes:
* Removed the deploy artifacts from the main branch workflow
* Deploy artifacts is now a manual workflow ([Docs can be found here](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow))
* Removed the git tag check for the deployment stage -- we won't want to use this as a trigger since github has yet to implement protected tags like they have for branches. ([See feature request thread here.](https://github.community/t/feature-request-protected-tags/1742)) 
* Addressed some dependency clashes that popped up

### DIFF
--- a/.github/workflows/deploy-artifacts.yml
+++ b/.github/workflows/deploy-artifacts.yml
@@ -1,7 +1,7 @@
 name: Deploy AaC Artifacts
 
 on:
-  workflow_call:
+  workflow_dispatch:
     # secrets:
     #   PYPI_API_TOKEN:
     #     required: true
--- a/.github/workflows/deploy-artifacts.yml
+++ b/.github/workflows/deploy-artifacts.yml
@@ -2,11 +2,11 @@ name: Deploy AaC Artifacts
 
 on:
   workflow_dispatch:
-    # secrets:
-    #   PYPI_API_TOKEN:
-    #     required: true
 
 jobs:
+  Build_Package_and_Artifacts:
+    uses: jondavid-black/AaC/.github/workflows/build-artifacts.yml@1e3f89c23eb6583116848d81ccd5ad952f9a62b6
+
   Deploy-Artifacts:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/deploy-artifacts.yml
```diff
@@ -27,7 +27,6 @@ jobs:
 
       - name: Publish distribution to PyPI
         run: echo "This is where I'd deploy if I were setup"
```
jondavid-black, line 29: Let's clean up the echo log entry to say "Attempting to publish to PyPI"

jondavid-black (commented):
We need a way to ensure version information is correctly handled when publishing.  I didn't see that in the PR...maybe I just missed it.

jondavid-black (approved):
I see how version works now.  We just need to ensure we manage the version in setup.py with AC on each issue.

.github/workflows/deploy-artifacts.yml
Coffee2Bits, line 29:
```suggestion
```

jondavid-black (approved):
Thanks for fixing the echo.

PR #178, opened 2022-01-06
147 Handle plugin execution results consistently

cvchaparro:
Closes: #147

Changes:
- Add `PluginExecutionResult` to allow handling execution results consistently.
- Update the CLI to display plugin messages and exit appropriately.
- Update `validation` context manager to return a `ValidationResult` structured object instead of returning the model or None, if the model is invalid.
- Add a way to allow users to handle returning `PluginExecutionResult`s and consistently handle various erroneous situations.
- Add a `PluginError` base class on which all plugin exception classes will be based
- Add an `OperationCancelled` exception that can be thrown if an operation is cancelled.
- Fix some linting issues in docstrings, etc.


jondavid-black (approved):
Looks good.  I notice there are some cases where only a success result is possible.  What happens if there is an unhappy path encountered...or is the happy path the only possibility?

Coffee2Bits:
> Looks good. I notice there are some cases where only a success result is possible. What happens if there is an unhappy path encountered...or is the happy path the only possibility?

I agree with

src/aac/plugins/gen_plugin/gen_plugin_impl.py
```diff
@@ -29,16 +35,31 @@ def generate_plugin(architecture_file: str) -> None:
...
+    result.set_messages(
+        f"Canceled: Move {architecture_file} to the desired directory and retry."
+    )
+
+    return result
+
 
+def _generate_plugin(architecture_file: str, plug_dir: str) -> PluginExecutionResult:
+    result = PluginExecutionResult(plugin_name, PluginExecutionStatusCode.SUCCESS)
+    result.set_messages(f"successfully created plugin in {plug_dir}")
```
Coffee2Bits, line 48:
```suggestion
    result.set_messages(f"Successfully created plugin in {plug_dir}")
```

src/aac/plugins/gen_plugin/gen_plugin_impl.py
```diff
@@ -29,16 +35,31 @@ def generate_plugin(architecture_file: str) -> None:
...
 
-def _generate_plugin(architecture_file: str, plug_dir: str) -> None:
     try:
         with validation(parser.parse_file, architecture_file) as parsed_model:
             templates = list(_compile_templates(parsed_model).values())
             write_generated_templates_to_file(templates, plug_dir)
     except GeneratePluginException as exception:
-        print(f"gen-plugin error [{architecture_file}]:  {exception}.")
+        result.status_code = PluginExecutionStatusCode.PLUGIN_FAILURE
+        result.set_messages(f"error [{architecture_file}]:  {exception}.")
```
Coffee2Bits, line 56:
```suggestion
        result.set_messages(f"Error [{architecture_file}]:  {exception}.")
```

src/aac/plugins/plugin_execution.py
```diff
@@ -0,0 +1,44 @@
...
+        """Add a message to the list of messages."""
+        self.messages.append(message)
+
+    def set_messages(self, *messages) -> None:
+        """Clear the current messages and set them to the passed in messages."""
+        self.messages = list(messages)
+
+    def success(self) -> bool:
+        """Return True if the command succeeded; False, otherwise."""
+        self.status_code == PluginExecutionStatusCode.SUCCESS
```
Coffee2Bits, line 42:
```suggestion
        return self.status_code == PluginExecutionStatusCode.SUCCESS
```

tests/plugins/test_gen_design_doc_impl.py
```diff
@@ -13,7 +14,8 @@ def test_can_generate_design_doc_with_models(self):
             with open(test_model_file_name, "w") as arch_file:
                 arch_file.write(TEST_MODEL)
 
-            gen_design_doc(test_model_file_name, temp_dir)
+            result = gen_design_doc(test_model_file_name, temp_dir)
+            self.assertEqual(result.status_code, PluginExecutionStatusCode.SUCCESS)
```
Coffee2Bits, line 18: Should add plugin failure/error handling. There could be IO issues that cause the plugin to fail.

tests/plugins/test_gen_protobuf.py
```diff
@@ -76,6 +82,16 @@ def test_gen_protobuf(self):
                 self.assertIn("TYPE_2", message_type_proto_file_contents)
                 self.assertIn("TYPE_3", message_type_proto_file_contents)
 
+    @patch('aac.plugins.gen_protobuf.gen_protobuf_impl.load_default_templates')
```
Coffee2Bits, line 85: Good find! I didn't know this decorator existed.

src/aac/plugins/gen_design_doc/gen_design_doc_impl.py
```diff
@@ -47,7 +54,10 @@ def gen_design_doc(architecture_files: str, output_directory: str, template_file
     )
     write_generated_templates_to_file([generated_document], output_directory)
 
-    print(f"Wrote system design document to {os.path.join(output_directory, output_filespec)}")
+    result = PluginExecutionResult(plugin_name, PluginExecutionStatusCode.SUCCESS)
+    result.add_message(f"Wrote system design document to {os.path.join(output_directory, output_filespec)}")
+
+    return result
```
Coffee2Bits, line 60: Should add error handling, there are potential IO issues that could cause plugin failure.

src/aac/plugins/gen_gherkin_behaviors/gen_gherkin_behaviors_impl.py
```diff
@@ -31,12 +36,16 @@ def gen_gherkin_behaviors(architecture_file: str, output_directory: str) -> None
         )
 
         write_generated_templates_to_file(generated_template_messages, output_directory)
-        print(f"Successfully generated templates to directory: {output_directory}")
+
+        result = PluginExecutionResult(plugin_name, PluginExecutionStatusCode.SUCCESS)
+        result.add_message(f"Successfully generated templates to directory: {output_directory}")
+
+        return result
```
Coffee2Bits, line 43: Should add error handling for potential errors like IO and validation failure.

src/aac/plugins/gen_json/gen_json_impl.py
```diff
@@ -1,18 +1,23 @@
...
     for architecture_file in architecture_files:
-        print(f"File: {architecture_file}")
         with validation(parse_file, architecture_file) as parsed_model:
-            _print_parsed_model(parsed_model)
-
+            messages.append(f"File: {architecture_file}\n{json.dumps(parsed_model)}")
 
-def _print_parsed_model(parsed_model: dict[str, any]) -> None:
-    print(json.dumps(parsed_model))
+    return PluginExecutionResult(plugin_name, PluginExecutionStatusCode.SUCCESS, messages)
```
Coffee2Bits, line 23: Another comment about handling error situations like being passed an invalid model.

Coffee2Bits (changes requested):
Looks good overall! 

I think there could be some improvements for catching general unexpected mayhem with the plugin results by leveraging a try/catch/finally block pattern that would ensure that despite whatever issues might happen in the plugin's execution won't bring down the application overall. This could be part of a larger error handling convention conversation, but I think it would be fair to effectively wrap all plugin executions in try/catch/block.

cvchaparro:
> Looks good. I notice there are some cases where only a success result is possible. What happens if there is an unhappy path encountered...or is the happy path the only possibility?

This is a good question that I came across, as well. (When you left this, I hadn't come up with a good solution to it but, now, I have a working solution that covers some failure scenarios - see below).

I think that, at least for some plugins, as long as the architecture file has a valid model, the only possibility is a "happy path" result (assuming no filesystem errors, etc., of course). If we did have filesystem or other issues - e.g. architecture file is not found, or whatever, then that should also cause an error result.

All of that said, I think an invalid model would (at a minimum) still constitute an unhappy path but it would require that we yield a different result from the `validation` context manager.

As a result of the above (and @Coffee2Bits's review) I went ahead and changed the validator (see [here](https://github.com/jondavid-black/AaC/pull/178#issue-1094852238) for details of what, specifically changed) and handled some obvious error scenarios.

src/aac/plugins/gen_plugin/gen_plugin_impl.py
```diff
@@ -15,30 +15,38 @@
...
 
-def _generate_plugin(architecture_file: str, plug_dir: str) -> None:
-    try:
-        with validation(parser.parse_file, architecture_file) as parsed_model:
-            templates = list(_compile_templates(parsed_model).values())
-            write_generated_templates_to_file(templates, plug_dir)
-    except GeneratePluginException as exception:
-        print(f"gen-plugin error [{architecture_file}]:  {exception}.")
+
+def _generate_plugin(architecture_file: str, plug_dir: str) -> PluginExecutionResult:
```
cvchaparro, line 45:
```suggestion
def _generate_plugin(architecture_file: str, plug_dir: str) -> str:
```

src/aac/validator.py
```diff
@@ -19,6 +19,20 @@ class ValidationError(RuntimeError):
     pass
 
 
+@attrs(slots=True, auto_attribs=True)
+class ValidationResult:
+    """Represents the result of validating a model.
+
+    Attributes:
+        is_valid (bool): Determines whether the model is valid, or not.
```
cvchaparro, line 27:
```suggestion
```

docs/old/Plugins.md
```diff
@@ -45,13 +45,13 @@ from aac.parser import parse_file, parse_str
 from aac.validator import validation
 
 def my_plugin_command(architecture_file: str):
-    with validation(parse_file, architecture_file) as model:
-        if not model:
+    with validation(parse_file, architecture_file) as result:
+        if result.is_valid:
```
Coffee2Bits, line 49: Is this not checking for an invalid result?
```suggestion
        if not result.is_valid:
```

Coffee2Bits (changes requested):
Great job! Just the one question on the readme content.

docs/old/Plugins.md
cvchaparro: Good catch! This is actually out-dated, as written. Updated in commit d3d79abf99297a454c6e78398d5373f7f55de1bb

Coffee2Bits (approved):
Looks good! 👍

PR #185, opened 2022-01-14
Add os matrix functional test

Coffee2Bits:
Over the winter break I got a curious hair and wanted to see how difficult it would be to run our functional tests using the OS matrix feature in Github Actions. Well, it turns out it worked! So, I've brought the changes over here. 

* Added functional tests for windows and macosx environments.

I had to replace a lot of the shell-based steps since Windows doesn't have standard shell terminals.

Only downside I see is that the extra testing will eat into our monthly "allowance" of Github Actions compute time.

### DIFF
--- a/.github/workflows/deploy-artifacts.yml
+++ b/.github/workflows/deploy-artifacts.yml
@@ -4,12 +4,12 @@ on:
   workflow_dispatch:
 
 jobs:
-  Build_Package_and_Artifacts:
+  Build_Artifacts:
     uses: jondavid-black/AaC/.github/workflows/build-artifacts.yml@1e3f89c23eb6583116848d81ccd5ad952f9a62b6
 
   Deploy-Artifacts:
     runs-on: ubuntu-latest
-    needs: Build_Package_and_Artifacts
+    needs: Build_Artifacts
     steps:
       - name: Download API Documentation
         uses: actions/download-artifact@v2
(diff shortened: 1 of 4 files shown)


jondavid-black:
From the GitHub docs...
`
Jobs that run on Windows and macOS runners that GitHub hosts consume minutes at 2 and 10 times the rate that jobs on Linux runners consume. For example, using 1,000 Windows minutes would consume 2,000 of the minutes included in your account. Using 1,000 macOS minutes, would consume 10,000 minutes included in your account.
`
So...having no idea what our current usage is, I'd suggest we remove macOS from the list.  In our near term target user community, there won't be any macOS.

Coffee2Bits:
> From the GitHub docs...
> 
> > Jobs that run on Windows and macOS runners that GitHub hosts consume minutes at 2 and 10 times the rate that jobs on
> > Linux runners consume. For example, using 1,000 Windows minutes would consume 2,000 of the minutes included in your
> > account. Using 1,000 macOS minutes, would consume 10,000 minutes included in your account.
> 
> So...having no idea what our current usage is, I'd suggest we remove macOS from the list. In our near term target user community, there won't be any macOS.

Wow! I believe Cameron works in a Mac environment so he can be the smoke test for now. I've removed macos from the OS matrix.

jondavid-black (approved):
Looks good to me.

PR #195, opened 2022-02-01
Update Gen Json

BryanBailey11:
updated plugin and developed automated tests for the plugin. I ran into a couple bugs regarding error reporting with a third test that has to be addressed in a future issue.


src/aac/plugins/gen_json/gen_json_impl.py
```diff
@@ -10,22 +11,51 @@
...
+    def to_json(architecture_file: str, output_directory: str) -> str:
+        file_name = os.path.basename(architecture_file)
+        file_ext = os.path.splitext(file_name)[1]
 
-    def to_json(architecture_file: str) -> str:
         with validation(parse_file, architecture_file) as result:
-            return f"File: {architecture_file}\n{json.dumps(result.model, indent=2)}\n"
+
+            if output_directory is None:
+                return f"File: {architecture_file}\n{json.dumps(result.model, indent=2)}\n"
```
jondavid-black, line 33: Should we remove the `File: {architecture_file}\n` portion of the output?  I'm imagining a scenario where someone needs the JSON version for use in a pipeline or something and wouldn't want to have to parse out that first line to use the output.

tests/plugins/test_gen_json.py
```diff
@@ -0,0 +1,150 @@
...
+          type: DataB
+      acceptance:
+        - scenario: go
+          given:
+            - ServiceOne is running.
+          when:
+            - The user sends a DataA request
+          then:
+            - The user receives a DataB response
+model:
```
jondavid-black, line 79: Missing the `---` separator between models.

tests/plugins/test_gen_json.py
```diff
@@ -0,0 +1,150 @@
...
+          type: DataC
+      acceptance:
+        - scenario: go
+          given:
+            - The System is running.
+          when:
+            - The System receives input.in
+          then:
+            - The System responds with output.out
+model:
```
jondavid-black, line 60: Missing the `---` separator between models.

jondavid-black:
I didn't do a full review but I did drop in some comments for you to look at.

src/aac/plugins/gen_json/gen_json_impl.py
```diff
@@ -10,22 +11,51 @@
...
             if not result.is_success():
                 status = result.status_code
 
     return PluginExecutionResult(plugin_name, status, messages)
+
+
+
+
+
+
```
cvchaparro, line 61: Delete the extra lines at the end of the file.

```suggestion
```

src/aac/plugins/gen_json/gen_json_impl.py
```diff
@@ -10,22 +11,51 @@
...
+                if file_ext == '':
+                    file_name = (file_name + '.json')
+                else:
+                    file_name = file_name.replace(file_ext, '.json')
+
+                output_file_path = os.path.join(output_directory, file_name)
+                outFile = open(output_file_path, "w")
+                outFile.write(json.dumps(result.model, indent=2))
+                outFile.close()
+                return f"Successfully added {file_name} to directory"
```
cvchaparro, line 45:
```suggestion
                return f"Successfully wrote JSON to {output_file_path}."
```

src/aac/plugins/gen_json/gen_json_impl.py
cvchaparro, line 33: Since we're using the same `json.dumps(...)` call in two places, now, it is a good candidate for being put into a variable.

cvchaparro (changes requested):

src/aac/plugins/gen_json/gen_json_impl.py
Coffee2Bits: Users can leverage the file output option if they absolutely needed a pipeline solution like that right now, but I think the larger question may be whether or not we want to add in some verbosity configuration that addresses this question for more than just this plugin. 

I think we could modify the existing plugin execution and execution result flow with a clearer delineation between the command output and any success/status messages. We could leverage this by exposing a general argument, like `-q, --quiet` that silences any output that isn't the return value of the command.

cvchaparro: I'd be up for that.

src/aac/plugins/gen_json/gen_json.yaml
```diff
@@ -10,11 +10,15 @@ model:
           type: string
           python_type: str
           description: filepath to the architecture file to convert to JSON.
+        - name: --output_directory
+          type: string
+          python_type: str
+          description: Directory pointing to the desired output for the JSON files (optional).
```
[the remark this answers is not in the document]
BryanBailey11: Thanks, I will keep that in mind.

tests/plugins/test_gen_json.py
```diff
@@ -0,0 +1,150 @@
...
+
+        with TemporaryDirectory(), NamedTemporaryFile("w") as temp_arch_file:
+            temp_arch_file.write(TEST_ARCH_YAML_STRING)
+            temp_arch_file.seek(0)
+
+            with plugin_result("", print_json, [temp_arch_file.name]) as result:
+                self.assertEqual(result.status_code, PluginExecutionStatusCode.SUCCESS)
+
+    def test_gen_json_with_invalid_arch_file(self):
+        pass
```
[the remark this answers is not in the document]
BryanBailey11: unfortunately not, The message still responds with success. Alex and I worked through that test yesterday, but were unable to get a failure to be raised to the test. The plugin execution status code still returned success even with an invalid YAML string. We also found a set of bugs within the validator.

src/aac/plugins/gen_json/gen_json_impl.py
```diff
@@ -10,22 +11,51 @@
...
+    """Prints the parsed_models from the parsed architecture_files values in JSON format.
+
+       (optional) if output directory is given to the cli, parsed_models with be writen to
+        a new file in user defined directory
+
+    """
+
+    def to_json(architecture_file: str, output_directory: str) -> str:
+        file_name = os.path.basename(architecture_file)
+        file_ext = os.path.splitext(file_name)[1]
```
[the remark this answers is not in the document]
Coffee2Bits: +1 It does look cleaner

tests/plugins/test_gen_json.py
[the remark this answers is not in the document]
Coffee2Bits: Cameron, the solution was non-obvious. It looks like there may be an issue with the exception propagating, but validation exceptions aren't being caught in the plugin result so even if validation fails the result status is still successful. I'll be taking a look today or tomorrow at it.

src/aac/plugins/gen_json/gen_json_impl.py
```diff
@@ -10,22 +11,51 @@
...
 plugin_name = "gen-json"
 
 
-def print_json(architecture_files: list[str]) -> PluginExecutionResult:
-    """Print the parsed_models from the parsed architecture_files values in JSON format."""
+def print_json(architecture_files: list[str], output_directory: str = None) -> PluginExecutionResult:
+    """Prints the parsed_models from the parsed architecture_files values in JSON format.
+
+       (optional) if output directory is given to the cli, parsed_models with be writen to
+        a new file in user defined directory
```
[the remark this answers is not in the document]
BryanBailey11: okay I will look into that format for future reference. Thanks for the correction

src/aac/plugins/gen_json/gen_json_impl.py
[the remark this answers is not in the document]
BryanBailey11: had to look this up, never used it before. Based on some research it is also much faster.

src/aac/plugins/gen_json/gen_json_impl.py
```diff
@@ -10,22 +11,51 @@
...
-            return f"File: {architecture_file}\n{json.dumps(result.model, indent=2)}\n"
+
+            if output_directory is None:
+                return f"File: {architecture_file}\n{json.dumps(result.model, indent=2)}\n"
+
+            else:
+                if file_ext == '':
+                    file_name = (file_name + '.json')
+                else:
+                    file_name = file_name.replace(file_ext, '.json')
```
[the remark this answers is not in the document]
BryanBailey11: How do you set the conditions of the if statement like in line 37 in this new code?

src/aac/plugins/gen_json/gen_json_impl.py
```diff
@@ -10,22 +11,51 @@
...
+
+                output_file_path = os.path.join(output_directory, file_name)
+                outFile = open(output_file_path, "w")
+                outFile.write(json.dumps(result.model, indent=2))
+                outFile.close()
```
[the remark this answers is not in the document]
BryanBailey11: agreed this is much cleaner.

src/aac/plugins/gen_json/gen_json_impl.py
BryanBailey11: can this be called within the same line? so it would look like this. 

export = json.dumps(result.model, indent=2)

return f"File: {architecture_file}\n{export}\n"

src/aac/plugins/gen_json/gen_json_impl.py
BryanBailey11: Since there might be a possibility of a list of files being parsed, I would like to add the file name to this line as well. I have made a commit with my fix.

src/aac/plugins/gen_json/gen_json_impl.py
BryanBailey11: fixed in new commit

src/aac/plugins/gen_json/gen_json_impl.py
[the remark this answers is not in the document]
Coffee2Bits: His suggestion is leveraging a [ternary operator](https://book.pythontips.com/en/latest/ternary_operators.html#ternary-operators) and ["truthy"](https://www.geeksforgeeks.org/truthy-vs-falsy-values-in-python/) values. So, `file_name.replace(file_ext, '.json')` is the block that executes if `file_ext` is "truthy" (not an empty string), otherwise filename is set to `f'{file_name}.json'`.

src/aac/plugins/gen_json/gen_json_impl.py
cvchaparro: The `file_name` will be included in the `output_file_path` - it'll just be an absolute file path rather than showing only the file name.

src/aac/plugins/gen_json/gen_json_impl.py
cvchaparro: I'm not really sure I understand what you mean by "called within the same line" since the code you provided was two lines. The only thing I'd add to what you put in your comment is that the `formatted_json = json.dumps(...)` part should go outside the `if output_directory` part so it can be used in the `out_file.write(...)` line.

jondavid-black: Me too...that's a great idea.

tests/plugins/test_gen_json.py
BryanBailey11: fixed with a code suggestion from Cam

tests/plugins/test_gen_json.py
BryanBailey11: fixed with a code suggestion from Cam

src/aac/plugins/gen_json/gen_json_impl.py
BryanBailey11: my bad, horrible wording. I was wondering if I could call the variable within the return line the same way json.dumps(...) is currently called. By your comment, I see that the answer is yes

src/aac/plugins/gen_json/gen_json_impl.py
BryanBailey11: okay, agreed and fixed with only output_file_path.

tests/plugins/test_gen_json.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+TEST_ARCH_YAML_STRING = """
```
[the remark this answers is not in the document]
BryanBailey11: Yaml String shortened to one model and three data types. The original was copied from test_gen_protobuf.py. We will want to remember to shorten it there as well.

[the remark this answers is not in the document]
BryanBailey11: shortened the YAML string. The original string was copied from test_gen_protobuf.py, might want to think about shortening it there as well.

src/aac/plugins/gen_json/gen_json_impl.py
```diff
@@ -10,22 +11,51 @@
...
+
+    def to_json(architecture_file: str, output_directory: str) -> str:
+        file_name = os.path.basename(architecture_file)
+        file_ext = os.path.splitext(file_name)[1]
 
-    def to_json(architecture_file: str) -> str:
         with validation(parse_file, architecture_file) as result:
-            return f"File: {architecture_file}\n{json.dumps(result.model, indent=2)}\n"
+
+            if output_directory is None:
```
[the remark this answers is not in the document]
BryanBailey11: updated the order to match this cleaner code structure

tests/plugins/test_gen_json.py
```diff
@@ -0,0 +1,48 @@
...
+    def test_print_json_with_output_directory(self):
+
+        with TemporaryDirectory() as temp_dir, NamedTemporaryFile("w") as temp_arch_file:
+            temp_arch_file.write(TEST_ARCH_YAML_STRING)
+            temp_arch_file.seek(0)
+
+            with plugin_result("", print_json, [temp_arch_file.name], temp_dir) as result:
+                print(result.messages)
+                print(result.status_code)
+                self.assertEqual(result.status_code, PluginExecutionStatusCode.SUCCESS)
```
cvchaparro, line 19: So I figured out what was going on here (and why we're getting that weird output. In tests, we don't want to use the `plugin_result` context manager (especially, here) because we add the return value of the command that's run (in this case `print_json`) which is expected to be a string to the list of messages. Instead, we can take the simple approach, as follows:

```suggestion
            result = print_json([temp_arch_file.name], temp_dir):
            # TODO: Delete these once testing is complete
            print(result.messages)
            print(result.status_code)
            self.assertEqual(result.status_code, PluginExecutionStatusCode.SUCCESS)
            self.assertIn("Wrote JSON to", "\n".join(result.messages))
```

tests/plugins/test_gen_json.py
```diff
@@ -0,0 +1,48 @@
...
+    def test_gen_json_output_to_cli(self):
+
+        with TemporaryDirectory(), NamedTemporaryFile("w") as temp_arch_file:
+            temp_arch_file.write(TEST_ARCH_YAML_STRING)
+            temp_arch_file.seek(0)
+
+            with plugin_result("", print_json, [temp_arch_file.name]) as result:
+                print(result.messages)
+                print(result.status_code)
+                self.assertEqual(result.status_code, PluginExecutionStatusCode.SUCCESS)
```
cvchaparro, line 30: See comment above for why we want to get the result like this:

```suggestion
            result = print_json([temp_arch_file.name])
            # TODO: Delete these once testing is complete
            print(result.messages)
            print(result.status_code)
            self.assertEqual(result.status_code, PluginExecutionStatusCode.SUCCESS)
            self.assertIn('"name": "Test_model"', "\n".join(result.messages))
```

tests/plugins/test_gen_json.py
```diff
@@ -0,0 +1,150 @@
...
+from unittest import TestCase
+from tempfile import TemporaryDirectory, NamedTemporaryFile
+
+from aac.plugins.gen_json.gen_json_impl import print_json
+from aac.plugins.plugin_execution import PluginExecutionStatusCode, plugin_result
+
+
+class Testgen_json(TestCase):
+
+    def test_print_json_with_output_directory(self):
```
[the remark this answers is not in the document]
cvchaparro: This can be addressed with the changes recommended in [this](https://github.com/jondavid-black/AaC/pull/195/files#r798974772) comment (or similar changes).

tests/plugins/test_gen_json.py
```diff
@@ -0,0 +1,48 @@
+from unittest import TestCase
+from tempfile import TemporaryDirectory, NamedTemporaryFile
+
+from aac.plugins.gen_json.gen_json_impl import print_json
+from aac.plugins.plugin_execution import PluginExecutionStatusCode, plugin_result
```
cvchaparro, line 5:
```suggestion
from aac.plugins.plugin_execution import PluginExecutionStatusCode
```

src/aac/plugins/gen_json/gen_json_impl.py
```diff
@@ -10,20 +10,42 @@
...
 
-def print_json(architecture_files: list[str]) -> PluginExecutionResult:
-    """Print the parsed_models from the parsed architecture_files values in JSON format."""
+def print_json(architecture_files: list[str], output_directory: str = None) -> PluginExecutionResult:
+    """Prints the parsed_models from the parsed architecture_files values in JSON format.
+
+       Arguments:
+           architecture_files (list[str]): filepath to the architecture file to convert to JSON.
+           output_directory (str): Directory in which JSON files will be written. (optional)
+
```
cvchaparro, line 23: `flake8` will throw pipeline failures if the spacing is not lined up with the ending `"""`.

```suggestion
    Arguments:
        architecture_files (list[str]): filepath to the architecture file to convert to JSON.
        output_directory (str): Directory in which JSON files will be written. (optional)
```

tests/plugins/test_gen_json.py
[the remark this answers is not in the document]
cvchaparro: The `test_gen_protobuf.py` YAML string may be needed for that because that's actually doing stuff with the different YAML "declarations" but here, we just want to see that the YAML gets converted to JSON, it doesn't matter much what the YAML actually is (as long as it validates).

python/tests/plugins/test_gen_json.py
```diff
@@ -0,0 +1,66 @@
+from unittest import TestCase
+from tempfile import TemporaryDirectory, NamedTemporaryFile
+
+from aac.plugins.gen_json.gen_json_impl import print_json
+from aac.plugins.plugin_execution import PluginExecutionStatusCode
+
+
+class Testgen_json(TestCase):
```
Coffee2Bits, line 8: [Python code style conventions](https://www.python.org/dev/peps/pep-0008/#class-names) dictate classes use CamelCase style. 
```suggestion
class TestGenJson(TestCase):
```

Coffee2Bits (changes requested):
Looks good, good work! Just a quick code convention to fix. I'm still working on the test issue and will update accordingly.

EDIT: The test has been fixed! 👍

cvchaparro (approved):
Nice work

PR #208, opened 2022-02-14
Add security scanning

jondavid-black:
Add security scanning for python and javascript.  This is using GitHub's default CodeQL scanner workflow.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'javascript', 'python' ]
+        language: [ 'typescript', 'python' ]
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript', 'python' ]
```
cvchaparro, line 35: We're actually doing the VSCode extension development using TypeScript.

```suggestion
        language: [ 'typescript', 'python' ]
```

Coffee2Bits (approved):
Looks good, great find!

PR #217, opened 2022-02-16
added workflow to export issues to markdown

jondavid-black:
This is an attempt to capture issues in a form that will help us work through some bureaucracy.  I took this from the gh2md repo where it seems to work fine, but I have no ability to test it.  Seems like a small risk because if the GH action fails we can choose to try and fix it or just delete the workflow it if is too painful.

### DIFF
--- a/.github/workflows/issues2md.yml
+++ b/.github/workflows/issues2md.yml
@@ -3,9 +3,7 @@
 # See https://github.com/mattduck/gh2md/issues/11.
 name: Issues2Markdown
 on:
-  schedule:
-    # every day
-    - cron: "0 0 * * *"
+  workflow_dispatch:
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/issues2md.yml
+++ b/.github/workflows/issues2md.yml
@@ -1,0 +1,38 @@
+# Action to backup issues using gh2md, contributed by
+# https://github.com/0ut0fcontrol.
+# See https://github.com/mattduck/gh2md/issues/11.
+name: Issues2Markdown
+on:
+  schedule:
+    # every day
+    - cron: "0 0 * * *"
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+      with:
+        persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal token
+        fetch-depth: 0 # otherwise, you will failed to push refs to dest repo
+    - name: Backup github issues to a markdown file.
+      run: |
+        pip3 install --user --upgrade setuptools
+        pip3 install --user gh2md
+        $HOME/.local/bin/gh2md $GITHUB_REPOSITORY issues.md --idempotent
+        git add issues.md
+      env:
+        GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Commit files
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
+        git diff --quiet && git diff --staged --quiet || git commit -am "Backup all issues into issues.md"
+    - name: Extract branch name
+      shell: bash
+      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+      id: extract_branch
+    - name: Push changes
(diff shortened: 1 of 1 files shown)


.github/workflows/issues2md.yml
```diff
@@ -0,0 +1,38 @@
...
+        git diff --quiet && git diff --staged --quiet || git commit -am "Backup all issues into issues.md"
+    - name: Extract branch name
+      shell: bash
+      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+      id: extract_branch
+    - name: Push changes
+      uses: ad-m/github-push-action@master
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        branch: ${{ steps.extract_branch.outputs.branch }}
```
Coffee2Bits, line 38: From my understanding, this will end up attempting to push to master which won't work. We could use the artifacts upload action to host the issues.md as a pipeline artifact rather than recommitting it into the repo.

jondavid-black: That would work for me...but I'll have to do some research on the artifacts upload action.

Coffee2Bits: Here's an example of us using it in the build artifacts workflow. 
https://github.com/jondavid-black/AaC/blob/b3d8881f2330fd5a5713f469abcd09eab4845ac4/.github/workflows/python-build-artifacts.yml#L56-L60

I think you'll replace that last stage with something like:
```
      - name: Upload Issues Markdwon
        uses: actions/upload-artifact@v2
        with:
          name: issues.md
          path: issues.md
```

.github/workflows/issues2md.yml
```diff
@@ -0,0 +1,29 @@
+# Action to backup issues using gh2md, contributed by
+# https://github.com/0ut0fcontrol.
+# See https://github.com/mattduck/gh2md/issues/11.
+name: Issues2Markdown
+on:
+  schedule:
```
Coffee2Bits, line 6: Is there a benefit to also making this a manually-started workflow, like the deployment workflow?

.github/workflows/issues2md.yml
jondavid-black: Done!  I removed all the commit, branch, and push commands.  Now it should just do an artifact upload.

.github/workflows/issues2md.yml
jondavid-black: Good point.  Another item for me to google really quick.

jondavid-black: OK...committed that change as well.  Now I can just run it on-demand.

.github/workflows/issues2md.yml
```diff
@@ -0,0 +1,27 @@
...
+        pip3 install --user gh2md
+        $HOME/.local/bin/gh2md $GITHUB_REPOSITORY issues.md --idempotent
+        git add issues.md
+      env:
+        GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Archive issues export
+      uses: actions/upload-artifact@v2
+      with:
+        name: issues export
+        path: issues.md
```
Coffee2Bits, line 27: This workflow works in my forked repo. You can see the run and output here: https://github.com/Coffee2Bits/AaC/actions/runs/1856072006
```suggestion
name: Export Issues to Markdown
on:
  workflow_dispatch:

jobs:
  build:
    name: Export Issues to Markdown
    runs-on: ubuntu-latest
    steps:
    - name: Backup Github Issues to Markdown
      run: |
        pip3 install --user --upgrade setuptools
        pip3 install --user gh2md
        $HOME/.local/bin/gh2md $GITHUB_REPOSITORY issues.md --idempotent
      env:
        GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    - name: Archive Issues Report
      uses: actions/upload-artifact@v2
      with:
        name: issues_export
        path: issues.md
```

Coffee2Bits (changes requested):
I cleaned out some unnecessary components like checking out the repository. I've got a whole-file suggestion that'll give you a working version of the export workflow.

jondavid-black (commented):
Thank you for improving and testing this.

Coffee2Bits (approved):

PR #230, opened 2022-02-23
Update gen-json to create output directories if they do not exist

### LINKED ISSUE #204: Gen-JSON fails to write JSON to file if the output file's directory doesn't exist
Gen-JSON fails to write JSON to file if the output file's parent directory doesn't exist. We should handle this more gracefully by creating the missing directories, interactively asking the user if they would like us to generate the directories, or failing with an error message that the directory is missing and needs to be created. 

## Replication:
Open a new gitpod instance. In the python workspace terminal enter the following:
```aac gen-json model/flow/System.yaml --output_directory test```

The following output will appear:
```
(venv) gitpod /workspace/AaC/python (main) $ aac gen-json model/flow/System.yaml --output_directory test
gen-json: general_failure

No such file or directory: test/System.json
```

## AC:
TBD From this discussion

cvchaparro:
Closes #204 

Changes:
- In the `gen-json` plugin, create output non-existent directories before trying to write the output JSON.

### DIFF
--- a/python/src/aac/plugins/gen_json/gen_json_impl.py
+++ b/python/src/aac/plugins/gen_json/gen_json_impl.py
@@ -15,7 +15,7 @@ plugin_name = "gen-json"
 
 
 def print_json(architecture_files: list[str], output_directory: str = None) -> PluginExecutionResult:
-    """Prints the parsed_models from the parsed architecture_files values in JSON format.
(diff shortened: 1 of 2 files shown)


Coffee2Bits (approved):
Looks good to me!

PR #244, opened 2022-03-10
Add yaml metadata to parser

### LINKED ISSUE #240: Add yaml metadata to Parser output
In order to support the heavily contextual nature of LSP features, we need to add some contextual information to our parsed definitions. Each definition should include some information to support document-level references at first.

## AC:
- [x] A metadata structure is for parsed definitions
  - [x] Each parsed definition/model should be wrapped in an encompassing data structure that includes metadata
- [x] Each parsed definition/model data structure must include:
  - [x] The absolute path to the file containing the definition
  - [x] The first line of the definition
  - [x] The last line of the definition
  - [x] The parsed definition/model
- [x] The parser is updated to populate the metadata structures

## Notes:
The closest support I can find from PyYaml for correlating metadata to YAML content is to use YAML scanning and tokens which seems about as effective as writing our own logic to correlate definitions to YAML string positions.

cvchaparro:
Closes #240

Changes:
- TBD

### DIFF
--- a/python/src/aac/parser.py
+++ b/python/src/aac/parser.py
@@ -12,6 +12,8 @@ import yaml
 from attr import Factory, attrib, attrs, validators
 from yaml.parser import ParserError as YAMLParserError
 
+from aac.util import is_path_name
+
 
 @attrs
 class SourceLocation:
@@ -45,7 +47,21 @@ class Lexeme:
     value: str = attrib(validator=validators.instance_of(str))
 
 
-def parse(source: str) -> list[Lexeme]:
+@attrs
+class ParsedModel:
+    """A parsed Architecture-as-Code model.
+
+    Attributes:
+        lexemes (list[Lexeme]): A list of lexemes for each item in the parsed model.
+        model (dict): The parsed model.
+    """
+
+    content: str = attrib(validator=validators.instance_of(str))
+    lexemes: list[Lexeme] = attrib(default=Factory(list), validator=validators.instance_of(list))
+    model: dict = attrib(default=Factory(list), validator=validators.instance_of(dict))
+
+
+def parse(source: str) -> ParsedModel:
     """Parse the Architecture-as-Code (AaC) model(s) from the provided source.
 
     Args:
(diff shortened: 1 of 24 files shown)


python/src/aac/parser.py
```diff
@@ -8,11 +8,105 @@
...
+
+    line: int = attrib(validator=validators.instance_of(int))
+    column: int = attrib(validator=validators.instance_of(int))
+    position: int = attrib(validator=validators.instance_of(int))
+    span: int = attrib(validator=validators.instance_of(int))
+
+
+@attrs
+class Lexeme:
+    """A lexical unit for a parsed AaC object.
```
Coffee2Bits, line 37: If I'm not mistaken, then this might clarify the summary a little.
```suggestion
    """A lexical unit for a parsed AaC definition.
```

python/src/aac/parser.py
```diff
@@ -8,11 +8,105 @@
...
+        value (str): The value of the parsed object.
+    """
+
+    location: SourceLocation = attrib(validator=validators.instance_of(SourceLocation))
+    source: str = attrib(validator=validators.instance_of(str))
+    value: str = attrib(validator=validators.instance_of(str))
+
+
+@attrs
+class ParsedModel:
```
Coffee2Bits, line 51: So, this is circling back to our conversation about terminology. If we're wanting to use `definition` to describe all possible AaC entries and let `model` describe those definitions with `model:` as their root key, then we should change this name for clarity. 
```suggestion
class ParsedDefinition:
```

Coffee2Bits: There are obviously numerous references throughout that will need to be changed, but I don't want to nit-pick them until we align on the naming above.

python/src/aac/parser.py
```diff
@@ -8,11 +8,105 @@
...
+    content: str = attrib(validator=validators.instance_of(str))
+    lexemes: list[Lexeme] = attrib(default=Factory(list), validator=validators.instance_of(list))
+    model: dict = attrib(default=Factory(list), validator=validators.instance_of(dict))
+
+
+def parse(source: str) -> ParsedModel:
+    """Parse the Architecture-as-Code (AaC) model(s) from the provided source.
+
+    Args:
+        source (str): The location from which to extract the Architecture-as-Code content.
```
Coffee2Bits, line 69:
```suggestion
        source (str): Must be either a file path to an AaC yaml file or a string containing AaC definitions.
```

python/src/aac/parser.py
```diff
@@ -8,11 +8,105 @@
...
+    """Get the YAML contents from the provided source.
+
+    Args:
+        source (str): The source from which to get the YAML contents. This can be a path-like
+        string pointing to a file with YAML contents; or a string of YAML contents.
+
+    Returns:
+        The YAML contents extracted from source.
+    """
+    if is_path_name(source):
```
Coffee2Bits, line 104: I think this makes it more explicit to the reader what is being tested. If is_path_name were more complex, then I'd agree with the abstraction. If `os.path.lexists` is too verbose `path.lexists` should work too.
```suggestion
    if os.path.lexists(source):
```

python/src/aac/util.py
```diff
@@ -155,3 +155,8 @@ def new_working_dir(directory):
     except Exception as e:
         chdir(current_dir)
         raise e
+
+
+def is_path_name(string: str) -> bool:
+    """Return True if string is a pathname; False, otherwise."""
+    return os.path.lexists(string)
```
Coffee2Bits, line 162: I would remove this; I don't think that the standard library name is too difficult to understand, and there isn't enough complexity abstracted away to warrant adding to `util.py`.  

I'm trying to keep the "utils" files slim, they're usually an indication that that codebase needs some organization and cleaning.

python/src/aac/parser.py
Coffee2Bits: Here's another thought: This change doesn't really allow us to give feedback if a user entered an incorrect file path. 

This check is assuming that if it's not a file that exists on the filesystem then it's a yaml string, which may not be the case. I think we could circumvent this issue if we did some argument validation, like plugins that list command arguments as being a `file` type would assert that those file arguments are in fact files that exist on the filesystem or kick back an error before even running the plugin command.

python/tox.ini
```diff
@@ -35,7 +35,7 @@ omit =
   src/aac/ui/*
 
 [report]
-fail_under = 91.75
+fail_under = 92.05
```
Coffee2Bits, line 38: I think we've reached a really good coverage at this point. Going for higher coverage will likely lead us into some bad and less useful testing practices. Personally, I think 90 is a good round number, but 92% is also good.
```suggestion
fail_under = 92.00
```

python/tests/helpers/io.py
```diff
@@ -8,6 +8,8 @@
 from contextlib import contextmanager
 from tempfile import NamedTemporaryFile, TemporaryDirectory
 
+from aac.util import new_working_dir
```
Coffee2Bits, line 11: I believe the only use of `new_working_dir` is here an in some other tests, we should move it from util.py to here.

Coffee2Bits (changes requested):
Looks good overall, but I did leave some feedback and some starters for potential discussions.

python/src/aac/parser.py
cvchaparro: I agree that we should do this. Perhaps we can create another story to handle this more thorough style of validation?

Coffee2Bits: Yeah, that would certainly be a separate story.

Coffee2Bits (approved):
Looks good. Good work!

PR #248, opened 2022-03-14
Implement Validator 2.0

### LINKED ISSUE #241: Add Validator Plugins
Add support for a new class of plugins, validators. 

## AC:
- [x] Expose an additional hookspec `validate(parsed_definition, **arguments)`
- [x] Add a new root key `validator` which describes the new validation definitions** that includes:
  - [x] name
  - [x] description
  - [x] behavior
- [x] Data definitions can reference validator definitions
  - [x] validator name
  - [x] additional validator arguments 

## Notes: 
** is an alternative to text-based matching
Behaviors should cover any future testing like generating BDD tests for validators.

### LINKED ISSUE #249: Validator 2.0 Part II
This story covers the general body of work for completing the switch to the new validation scheme.

## AC:
- [ ] Each definition that is validated is traversed and all applicable, registered validators are gathered
- [ ] All applicable validators are run on each definition
- [ ] All validator feedback is collected into a set of findings (info, errors, warnings)
- [ ] The following validators are created:
  - [ ] Defined references
  - [ ] Required fields

### LINKED ISSUE #255: Create the Defined References validation plugin
Create the "Defined References" validator plugin. This validator plugin verifies that referenced definitions are defined.
## AC:
- [ ] The validator plugin implements the following plugin hooks
  - [ ] `get_plugin_aac_definitions()`
  - [ ] `register_validators()`
- [ ] The validator contains acceptance scenarios for
  - [ ]  validation success
  - [ ]  validation failure
- [ ] Validation fails if a type field references a definition that doesn't exist in the active context

### LINKED ISSUE #256: Create the Required Fields validation plugin
Create the "Required Fields" validator plugin. This validator plugin verifies that every field declared in the `required` section is present and populated.

## AC:
  - [x] `register_validators()`
  - [x]  validation success
  - [x]  validation failure
- [x] Validation fails if a definition does not define and populate every required field

Coffee2Bits:
Closes #241
Closes #249
Closes #255
Closes #256

* Added a hook for validator plugins to register with
* Created a required fields validator plugin
* Created a type reference validator plugin
* Updated the core spec to support adding validators to `data` definitions
* Added support for validation arguments in the core spec
* Moved ParsedDefinition to Definition in the lang module
* Added language features for dealing with AaC definitions
* Added basic configuration for logging. New log `aac.log` is created locally.
* Changed ParsedDefinition/Definition structures to be 1:1 with YAML definitions, parsing now returns a list of Definitions
* Converted the old ValidationContext into a LanguageContext


python/src/aac/spec/spec.yaml
```diff
@@ -49,6 +49,16 @@ data:
...
+    - name: validation
+      type: Validation
+      description: |
+        A declaration for a validation definition.
+
+        Validation definitions must be paired with a corresponding validation
+        plugin. Validations are referenced by name in the definition that they
+        are applied to. For instance, the validation definition references a
+        validation plugin that ensures validation definitions have a corresponding
+        plugin capable of executing some validation check.
```
cvchaparro, line 61: Honestly, this "For instance"  part reads a bit strangely to me with references to validation definitions then plugins back to definitions then back to plugins. Perhaps we could reword a bit to specify more clearly why the validation definitions exist and what purpose they serve.

python/src/aac/plugins/validators/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    def is_validation_definition(definition):
+        return "validation" in definition
+
+    parsed_validator_definitions = parser.parse_str(source_name, plugin_definitions_string)
+    validation_definitions = list(filter(is_validation_definition, parsed_validator_definitions.values()))
+
+    if len(validation_definitions) != 1:
+        raise RuntimeError(
```
cvchaparro, line 25: I think we should use a specific error, here.

python/src/aac/validation/ValidatorPlugin.py
```diff
@@ -0,0 +1,37 @@
...
+@attrs
+class ValidatorPlugin:
+    """
+    A class that contains all the relevant information to manage and execute validator plugins with the validation process.
+
+    Attributes:
+        name: A string with the name of the command argument
+        validation_definition: The AaC definition of the Validator Plugin
+        validation_function: The validation callback function
+        validation_arguments: Any additional arguments to supply to the validation_function
```
cvchaparro, line 15: Is there a need to have the `validation_` prefix, here? Since it's part of a `ValidatorPlugin`, I'd say this is not really helpful information.

python/src/aac/validation/ValidatorPlugin.py
```diff
@@ -0,0 +1,37 @@
+"""ValidatorPlugin class."""
+from __future__ import annotations
```
cvchaparro, line 2: Can we delete this? It doesn't seem to be used anywhere.

```suggestion
```

python/tests/plugins/validators/test_validators.py
```diff
@@ -0,0 +1,19 @@
+from unittest import TestCase
+
+from aac.plugins.validators import get_validation_definition_from_plugin_definitions
+
+
+class TestValidationModule(TestCase):
```
cvchaparro, line 6: I think that this class having the same name as the one in `python/tests/plugins/validators/test_validator_implementation.py` could cause problems testing. Can you confirm both tests are actually being run?

python/tests/plugins/validators/test_validators.py
```diff
@@ -0,0 +1,19 @@
...
+    def test_get_validation_definition_from_plugin_definitions(self):
+        parsed_definition = get_validation_definition_from_plugin_definitions(__name__, TEST_VALIDATION_DEFINITION_STRING)
+        definition = parsed_definition.get("validation")
+        definition_name = definition.get("name")
+        self.assertEqual(TEST_VALIDATION_NAME, definition_name)
```
cvchaparro, line 11: Is it currently the case that validation functions are not being used? If they are, how are they being defined?

cvchaparro (changes requested):
Nice work, overall, though I do have a few questions on how this is working.

Coffee2Bits:
> Nice work, overall, though I do have a few questions on how this is working.

Yeah, this isn't 100%. I have to follow through with some more work for the validator overhaul this upcoming sprint. I'll bring it up in stand-up.

Coffee2Bits:
I'm pulling this back into a draft; I'm not comfortable committing a partial solution. We'll just extend this branch a little as we round out the validator overhaul.

python/src/aac/validation/ValidatorPlugin.py
Coffee2Bits: The future annotation is necessary for the type hinting on line 24.

python/src/aac/plugins/validators/validator_implementation/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+         string representing yaml extensions and data definitions employed by the plugin
+    """
+
+    return get_resource_file_contents(__package__, PLUGIN_YAML_FILE)
+
+
+@hookimpl
+def register_validators() -> ValidatorPlugin:
+    """
+    Returns the information about the validation plugin necessary to execute validation.
```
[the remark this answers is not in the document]
Coffee2Bits: Good call out.

python/tests/plugins/validators/test_validators.py
Coffee2Bits: They do. Thanks for spotting that, copy mistake. :x

python/src/aac/cli.py
```diff
@@ -13,6 +13,7 @@
 from aac.plugins.plugin_execution import PluginExecutionResult, plugin_result
 from aac.spec.core import get_aac_spec_as_yaml
 from aac.validator import validation
+from aac.validation import preview_validation
```
Coffee2Bits, line 16: Remove the preview validation command before moving this PR out of draft.

.vscode/python.code-workspace
```diff
@@ -72,6 +76,18 @@
...
+            "type": "python",
+            "request": "launch",
+            "program": "${workspaceFolder}/src/aac/__main__.py",
+            "args": [
+                "validate-preview",
+                "${workspaceFolder}/src/aac/plugins/validators/validation_module/validation_module.yaml"
+            ],
+            "console": "integratedTerminal",
+            "justMyCode": false
+        },
```
Coffee2Bits, line 90: Remove this.
```suggestion
```

python/src/aac/lang/__init__.py
```diff
@@ -1 +1,9 @@
 """A module that provides AaC language features."""
+
+from ._active_context import ActiveContext
+from .context_manager import get_active_context
```
Coffee2Bits, line 4: I think if we expose a submodule in a package, then the submodule should be made private and treated as such.

python/tests/lang/test_references.py
```diff
@@ -0,0 +1,33 @@
+from unittest import TestCase
+
+from aac.lang.references import get_definition_references_from_list
+
+from tests.helpers.parsed_definitions import create_data_definition, create_data_ext_definition, create_field_entry, create_model_definition
+
+
+class TestLangReferences(TestCase):
+
+    def test_get_definition_references_from_list(self):
```
Coffee2Bits, line 10: Need to add a test to demonstrate that references won't return the source definition as a reference. In other words, the declaration of a definition is not a reference to itself.

python/src/aac/parser/_parsed_definition.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    name: str = attrib(validator=validators.instance_of(str))
+    content: str = attrib(validator=validators.instance_of(str))
+    lexemes: list[Lexeme] = attrib(default=Factory(list), validator=validators.instance_of(list))
+    definition: dict = attrib(default=Factory(list), validator=validators.instance_of(dict))
+
+    def get_root_key(self) -> str:
+        """Return the root key for the parsed definition."""
+        return list(self.definition.keys())[0]
```
Coffee2Bits, line 27: This should be able to handle a definition with imports. Make sure to test that.

python/src/aac/lang/definition_helpers.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+    ret_val = {}
+    for key, value in models.items():
+        if root_name in value.keys():
+            ret_val[key] = value
+
+    return ret_val
+
+
+def get_definitions_by_root_key(parsed_definitions: list[ParsedDefinition], root_key: str) -> list[ParsedDefinition]:
```
Coffee2Bits, line 170: For these definition helpers, the args should look like:
```suggestion
def get_definitions_by_root_key(root_key: str, parsed_definitions: list[ParsedDefinition]) -> list[ParsedDefinition]:
```

python/src/aac/lang/_active_context.py
```diff
@@ -0,0 +1,136 @@
...
+import copy
+
+from attr import Factory, attrib, attrs, validators
+
+from aac.parser import ParsedDefinition
+from aac.lang.definition_helpers import get_definition_by_name, get_definitions_by_root_key, search
+
+
+@attrs(slots=True, auto_attribs=True)
+class ActiveContext:
```
Coffee2Bits, line 12: Alternative naming could be like LanguageContext where ActiveContext specifically refers to the managed context instance.

python/src/aac/plugins/validators/__init__.py
Coffee2Bits: Moved the code, but made the runtime error a plugin one.

python/src/aac/lang/definition_helpers.py
```diff
@@ -0,0 +1,266 @@
+"""Arch-as-Code helper functions to simplify managing and working with parsed definitions/models."""
```
Coffee2Bits, line 1: Some consideration should go to splitting these out into more lang modules

python/src/aac/lang/_active_context.py
```diff
@@ -0,0 +1,165 @@
...
+from aac.lang.definition_helpers import (
+    get_definition_by_name,
+    get_definitions_by_root_key,
+    search_definition,
+    remove_list_type_indicator,
+)
+
+
+@attrs(slots=True, auto_attribs=True)
+class ActiveContext:
```
Coffee2Bits, line 15: Should the ActiveContext manage the list of actively-contributing plugins and sources? This would allow us to control the application of plugins at the context level which will have implications for things like tests.

python/src/aac/validate/_collect_validators.py
```diff
@@ -0,0 +1,99 @@
...
+                f"Failed to find applicable validator plugin for name '{validation_name}' out of the available plugins:\n{applicable_validation_dicts}"
+            )
+
+    return list(applicable_validators.values())
+
+
+def _get_validation_entries_from_definition_fields(definition: ParsedDefinition, context: ActiveContext) -> list[dict]:
+    """Return a list of validation entries from substructures and fields in the definition."""
+
+    def _recursively_gather_field_validations(field_definition: ParsedDefinition, traversed_types: list[str]) -> list[dict[str, list]]:
```
Coffee2Bits, line 41: Fix the recursion - recurse through the fields, not the definition.

python/src/aac/validate/_validate.py
```diff
@@ -0,0 +1,63 @@
...
+from aac.lang.context_manager import get_active_context
+from aac.parser import parse, ParsedDefinition
+from aac.plugins.plugin_manager import get_validator_plugins
+from aac.plugins.validators import ValidatorResult
+from aac.validate import ValidationResult, ValidationError
+from aac.validate._validate_definition import validate_definition
+
+
+@contextmanager
+def validate_definitions(user_definitions: list[ParsedDefinition]) -> ValidationResult:
```
cvchaparro, line 13: Two things:
1. It looks like we can extract most of this code into a separate function and just pass the different parameter, if needed, for `validate_source` and `validate_definitions`. Or, perhaps, `validate_source` can call `validate_definitions`.
2. I'd prefer naming these context managers something like `validated_source`/`validated_definitions`, etc since they are intended to be used from a `with` statement. `with <verb>` sounds a bit awkward.

python/src/aac/validate/_validate_definition.py
```diff
@@ -0,0 +1,123 @@
...
+from aac.lang.hierarchy import get_definition_ancestry
+from aac.parser import ParsedDefinition
+from aac.plugins.validators import ValidatorPlugin, ValidatorResult
+from aac.lang.context import ActiveContext
+
+
+def validate_definition(
+    definition: ParsedDefinition, validator_plugins: list[ValidatorPlugin], context: ActiveContext
+) -> list[ValidatorResult]:
+    """Traverse the definition validate it."""
```
cvchaparro, line 18:
```suggestion
    """Traverse the definition and validate it according to the validator plugins."""
```

python/src/aac/validate/_validate_definition.py
```diff
@@ -0,0 +1,123 @@
...
+                _apply_validation(fields_to_validate, applicable_validators)
+                # Need to recurse here for embedded definitions
+
+            # Else if the field is an array type
+            elif isinstance(fields_to_validate, list):
+                for field_to_validate in fields_to_validate:
+                    _apply_validation(field_to_validate, applicable_validators)
+                    # Need to recurse here for embedded definitions
+
+    # Pre-populate with the root key a being traversed since we have already validate the root structure
```
cvchaparro, line 83: I am not seeing where this comment is relevant to the line that follows.

python/src/aac/validate/_validate_definition.py
```diff
@@ -0,0 +1,123 @@
...
+            field_type_validations = _get_validation_entries(field_definition)
+            applicable_validators = [
+                _get_validator_plugin_by_name(validation.get("name"), validator_plugins)
+                for validation in field_type_validations
+            ]
+
+            # If the field is not an array type
+            if isinstance(fields_to_validate, dict):
+                _apply_validation(fields_to_validate, applicable_validators)
+                # Need to recurse here for embedded definitions
```
cvchaparro, line 75: Is this effectively a TODO comment? If so, could we make that into a story (if it's not already done) so we don't forget to do that?

python/src/aac/validate/_validate_definition.py
```diff
@@ -0,0 +1,123 @@
...
+
+            # Else if the field is an array type
+            elif isinstance(fields_to_validate, list):
```
cvchaparro, line 78: What's the `else` clause. What else could `fields_to_validate` be if not a list or a dict? If these are the only possibilities, could we just turn the `elif` into an `else`?

python/tests/helpers/test_io.py
```diff
@@ -0,0 +1,24 @@
...
+from unittest import TestCase
+from tempfile import TemporaryDirectory
+from tests.helpers.io import new_working_dir
+
+
+class TestArchUtil(TestCase):
+    """
+    Unit test class for aac.util module.
+    """
+
```
cvchaparro, line 15:
```suggestion
class TestIOHelpers(TestCase):
```

python/tests/helpers/test_io.py
```diff
@@ -0,0 +1,24 @@
...
+    def test_new_working_dir(self):
+        """
+        Unit test for the util.new_working_dir method.
+        """
```
cvchaparro, line 19: Are we going to start using docstrings in tests, now?

```suggestion
```

python/tests/helpers/test_io.py
```diff
@@ -0,0 +1,24 @@
...
+        original_working_directory = getcwd()
+
+        with TemporaryDirectory() as temp_directory, new_working_dir(temp_directory):
+            self.assertNotEqual(original_working_directory, temp_directory)
+            self.assertEqual(getcwd(), temp_directory)
```
cvchaparro, line 24: We should also check that outside of the context manager, the current working directory is back to `original_working_directory`.

```suggestion
            self.assertEqual(getcwd(), temp_directory)

        self.assertEqual(getcwd(), original_working_directory)
```

python/tests/plugins/test_gen_design_doc_impl.py
```diff
@@ -15,7 +16,7 @@ def test_can_generate_design_doc_with_models(self):
                 arch_file.write(TEST_MODEL)
 
             result = gen_design_doc(test_model_file_name, temp_dir)
-            self.assertEqual(result.status_code, PluginExecutionStatusCode.SUCCESS)
+            assert_plugin_success(result)
```
cvchaparro, line 19: I actually have something like this stashed but I'm glad you beat me to it. 😄

python/tests/validate/test_collect_validators.py
```diff
@@ -0,0 +1,114 @@
...
+
+        validation_plugins = get_validator_plugins()
+
+        self.assertGreater(len(validation_plugins), 0)
+
+        first_validator_name = validation_plugins[0].name
+        self.assertEqual(validation_plugins[0], _get_validator_plugin_by_name(first_validator_name, validation_plugins))
+
+        # second_validator_name = validation_plugins[1].name
+        # self.assertListEqual(validation_plugins[1], _get_validator_plugin_by_name(second_validator_name, validation_plugins))
```
cvchaparro, line 114: Why is this commented?

cvchaparro (changes requested):

python/src/aac/validate/_collect_validators.py
```diff
@@ -0,0 +1,99 @@
+import logging
```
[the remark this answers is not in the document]
Coffee2Bits: I'm doing a lot of refining around here; we can sync so you have an idea.

python/tests/validate/test_collect_validators.py
Coffee2Bits: It's waiting for me to implement the required fields validator plugin, then I'll uncomment it.

python/src/aac/validate/_validate.py
Coffee2Bits: To point 2. I was thinking along the lines of `with open(file) as fh:`, but I did some googling and apparently there is a PEP for this: https://peps.python.org/pep-0343/#examples

Though, I'll have to do some more reading to understand the implication of the `__enter__` vs `__exit__` disctinction.

python/src/aac/cli.py
```diff
@@ -65,8 +67,12 @@ def _validate_cmd(model_file: str) -> PluginExecutionResult:
...
 
     def validate_model() -> str:
-        with validation(parser.parse, model_file):
-            return f"{model_file} is valid"
+        user_definitions = parser.parse(model_file)
+        with validate_definitions(user_definitions) as validation_result:
+            if validation_result.is_valid:
+                return f"{model_file} is valid"
+            else:
+                raise ValidationError(f"Invalid definitions:\n{user_definitions}\nMessages:\n{validation_result.messages}")
```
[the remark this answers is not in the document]
Coffee2Bits: I've removed this implementation, the validation error is only being raised from within the validate context manager.

python/src/aac/lang/_active_context.py
```diff
@@ -0,0 +1,165 @@
...
+            self.add_definition_to_context(extension_definitions)
+
+    def get_root_keys(self) -> list[str]:
+        """
+        Get the list of root keys as defined in the ActiveContext.
+
+        Returns a list of strings indicating the current root keys in the active context. These
+        keys may differ from those provided by the core spec since the ActiveContext applies definitions
+        from active plugins and user files, which may extend the set of root keys.
+        See :py:func:`aac.spec.get_root_keys()` for the list of root keys provided by the unaltered core AaC DSL.
```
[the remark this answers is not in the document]
Coffee2Bits: Formatting keys? Are you referencing the docstring `:py:func:`aac.spec.get_root_keys()` section?

python/src/aac/spec/spec.yaml
```diff
@@ -216,9 +277,15 @@ data:
...
         definition.
+    - name: validation
+      type: ValidationReference[]
+      description: |
+        References and additional arguments for validations to apply to the definition.
   required:
     - name
     - fields
+  # validation:
+  #   - name: Required fields are present
```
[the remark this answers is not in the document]
Coffee2Bits: It will be applied in this PR.

python/tests/helpers/test_io.py
Coffee2Bits: It's vestigial. I don't want to.

python/src/aac/lang/context.py
```diff
@@ -0,0 +1,165 @@
+"""The Active Context maintains the highly-contextual AaC DSL."""
```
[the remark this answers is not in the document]
Coffee2Bits: I've moved the context to language_context. That being said, I don't quite understand the underlying concern, is this still something we need to discuss after the move to the `aac.lang.language_context.py`?

python/src/aac/lang/references.py
```diff
@@ -0,0 +1,35 @@
...
+from aac.parser import ParsedDefinition
+from aac.lang.context_manager import get_active_context
+
+
+def get_definition_references_from_context(source_definition: ParsedDefinition) -> list[ParsedDefinition]:
+    """
+    Return a list of ParsedDefinitions from the current ActiveContext that reference the source_definition.
+
+    Args:
+        source_definition: The definition that is being referenced
```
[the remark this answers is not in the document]
Coffee2Bits: Type changed, but I've updated the resulting docstring.

python/src/aac/lang/references.py
```diff
@@ -0,0 +1,35 @@
...
+        definitions_to_search: The definitions to search through for references
+
+    Return:
+        A list of ParsedDefinitions that reference the source_definition_name
+    """
+    def _filter_definitions_by_reference(definition_to_filter: ParsedDefinition) -> bool:
+        return f"type: {source_definition.name}" in definition_to_filter.to_yaml()
+
+    return list(filter(_filter_definitions_by_reference, definitions_to_search))
```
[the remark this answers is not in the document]
Coffee2Bits: Looks like I already addressed this in my local changes.

python/src/aac/lang/context_manager.py
```diff
@@ -0,0 +1,34 @@
+"""This module manages the ActiveContext instance and its lifecycle."""
```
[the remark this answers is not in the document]
Coffee2Bits: Changed to `active_context_lifecycle_manager` for now.

python/src/aac/lang/definition_helpers.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def does_definition_name_match(parsed_definition: ParsedDefinition) -> bool:
+        return remove_list_type_indicator(definition_name) == parsed_definition.name
+
+    matching_definitions = list(filter(does_definition_name_match, parsed_definitions))
+    matching_definitions_count = len(matching_definitions)
+
+    result = None
+    if matching_definitions_count > 1:
+        logging.error('Found multiple definitions with the same name "{definition_name}"\n{matching_definitions}"')
```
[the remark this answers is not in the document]
Coffee2Bits: Good catch, I seem to have done this a few times. Gotta hunt for this copy paste error/oversight.

python/src/aac/lang/definition_helpers.py
```diff
@@ -0,0 +1,266 @@
...
+
+def is_array_type(type: str) -> bool:
+    """Returns a boolean indicating if the field is an array of multiple entries."""
+    return type.endswith("[]")
+
+
+def get_definition_fields_and_types(
+    parsed_definition: ParsedDefinition, context_definitions: list[ParsedDefinition]
+) -> dict[str, ParsedDefinition]:
+    """Return a list of field names to their definition types."""
```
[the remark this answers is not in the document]
Coffee2Bits: So, I tried to avoid using the term "mapped" since python is out here trying to be all fancy with dictionaries instead of maps. I didn't want to inject my java bias, but if you also agree that "mapped" is a good description for dictionaries, then I'm happy to make the change.

python/src/aac/lang/hierarchy.py
```diff
@@ -0,0 +1,54 @@
...
+from aac.lang.context import ActiveContext
+from aac.lang.definition_helpers import get_definition_by_name, search_definition
+from aac.parser import ParsedDefinition
+
+
+def get_definition_ancestry(definition: ParsedDefinition, context: ActiveContext) -> list[ParsedDefinition]:
+    """Return an ordered list of ancestor definitions starting with the root definition and ending with the argument."""
+
+    ancestors_list = []
+    found_self_defined_ancestor = False
```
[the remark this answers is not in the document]
Coffee2Bits: Yes, currently only `data` is self-defining, but I'm trying to leave some flexibility in here in case someone elected to self-define another structure. That being said, we make some validation assumptions on the fact that `data:` is the base definition, so I haven't thought of a good way to be 100% assumption free.

python/src/aac/lang/references.py
```diff
@@ -0,0 +1,35 @@
...
+    Return a subset of ParsedDefinitions that reference the source_definition.
+
+    Args:
+        source_definition: The definition that is being referenced
+        definitions_to_search: The definitions to search through for references
+
+    Return:
+        A list of ParsedDefinitions that reference the source_definition_name
+    """
+    def _filter_definitions_by_reference(definition_to_filter: ParsedDefinition) -> bool:
```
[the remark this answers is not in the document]
Coffee2Bits: I prefer the underscore as it reinforces the idea that the function is non-public. I'd like to switch to inner functions starting with an underscore. I get that you can't access those inner functions from outside, but I prefer the ability to immediately determine whether or not a function is private/public.

python/tests/helpers/test_io.py
Coffee2Bits: Good call. :)

python/src/aac/lang/_active_context.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        Add the list of ParsedDefinitions to the list of definitions in the ActiveContext, any extensions are added last.
+
+        Args:
+            parsed_definitions: The list of ParsedDefinitions to add to the context.
+        """
+
+        extension_definitions = get_definitions_by_root_key("ext", parsed_definitions)
+
+        # Avoiding the use of sorted() since we already deepcopy each definition as we add it to the context.
```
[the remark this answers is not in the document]
Coffee2Bits: Vestigial from a previous approach. 🙃

python/src/aac/lang/_active_context.py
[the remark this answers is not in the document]
cvchaparro: Honestly I'm not sure why I called them "keys", but yes. I suggested using formatting markers (maybe?) and was told not to because it made the "docstrings less readable". Just curious if we are doubling back on that, now.

python/src/aac/lang/context.py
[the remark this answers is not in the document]
cvchaparro: The concern was that it was defined in two different files `_active_context.py` and `context.py`. If that's no longer the case, then this can be marked as resolved.

python/src/aac/lang/definition_helpers.py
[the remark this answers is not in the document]
cvchaparro: I don't think "mapped" implies a Java bias and I think most devs of any experience would understand the different terminology for the data structure, in Common Lisp we call them hash tables :shrug: - it's pretty much all the same thing regardless.

python/src/aac/lang/hierarchy.py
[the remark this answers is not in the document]
cvchaparro: I don't think we'll ever get 100% assumption-free. Although it would be nice, I think it's impractical - even most of what we know about math is based on a few [axioms](https://encyclopediaofmath.org/wiki/Axiomatic_method). I know the end goal is for the AaC language to be self-defining, etc but I think we can only ever approximate that.

python/src/aac/lang/references.py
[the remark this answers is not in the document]
cvchaparro: If it's an inner function, it's clearly non-public and the ability to immediately see that is provided by the indentation. The leading underscore for inner functions is ultimately just extra noise, in my opinion.

python/src/aac/lang/definitions/structure.py
```diff
@@ -0,0 +1,110 @@
...
+    if source_definition_root_key == sub_schema_definition.name:
+        substructure_instances = [source_definition_fields]
+    else:
+        source_definition_schema = get_definition_schema(source_definition, context)
+        _get_substructures(source_definition_schema, source_definition_fields)
+
+    return substructure_instances
+
+
+def get_sub_definitions(source_definition: Definition, context: LanguageContext) -> list[Definition]:
```
Coffee2Bits, line 74: Move to schema module?

python/src/aac/lang/_active_context.py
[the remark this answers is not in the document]
Coffee2Bits: Hmm. I don't see why that would be a big issue. The formatting here is in the long description, so it's not really impacting the high-value areas like summary, args, and return. 

Do you know where that feedback occurred in the other review? I'm curious if there is something about that particular instance that influenced that feedback.

python/src/aac/lang/language_context.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        Get the list of primitive types as defined in the LanguageContext.
+
+        Returns a list of strings indicating the currently defined primitive types in the active context.
+        These types may differ from those provided by the core spec since the LanguageContext applies definitions
+        from active plugins and user files, which may extend the set of root keys.
+        See :py:func:`aac.spec.get_primitives()` for the list of root keys provided by the unaltered core AaC DSL.
+
+        Returns:
+            A list of strings, one entry for each root name available in the LanguageContext.
```
Coffee2Bits, line 129:
```suggestion
        from active plugins and user files, which may extend the set of primitive types. See
        :py:func:`aac.spec.get_primitives()` for the list of primitive types provided by the unaltered core AaC DSL.
        Returns:
            A list of strings, one entry for each primitive type available in the LanguageContext.
```

python/src/aac/lang/_active_context.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        Get the list of primitive types as defined in the ActiveContext.
+
+        Returns a list of strings indicating the currently defined primitive types in the active context.
+        These types may differ from those provided by the core spec since the ActiveContext applies definitions
+        from active plugins and user files, which may extend the set of root keys.
+        See :py:func:`aac.spec.get_primitives()` for the list of root keys provided by the unaltered core AaC DSL.
+
+        Returns:
+            A list of strings, one entry for each root name available in the ActiveContext.
```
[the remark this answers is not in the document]
Coffee2Bits: Moved this suggestion to the new file.

python/src/aac/lang/language_context.py
```diff
@@ -0,0 +1,182 @@
...
+        return search_definition(self.get_primitives_definition(), ["enum", "values"])
+
+    def get_defined_types(self) -> list[str]:
+        """
+        Get the list of types defined by other definitions in the LanguageContext.
+
+        Returns a list of strings of all definition types in the active context.
+
+        Returns:
+            A list of strings, one entry for each definition name available in the LanguageContext.
```
Coffee2Bits, line 140:
```suggestion
        Return a list of strings containing the names of all the definitions in the LanguageContext.
        
        Returns:
            A list of strings, one entry for each definition name available in the LanguageContext.
```

python/src/aac/lang/hierarchy.py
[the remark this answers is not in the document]
Coffee2Bits: That's fair. I'll document this assumption as part of my validator 2.0 docmentation.

python/src/aac/lang/definitions/structure.py
Coffee2Bits: + rename to....? Maybe something about the schema definition's composition?

python/src/aac/lang/definition_helpers.py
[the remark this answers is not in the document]
Coffee2Bits: Agreed - I'll accept the suggestion.

python/src/aac/lang/references.py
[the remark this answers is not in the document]
Coffee2Bits: The definition of the inner function is obviously non-public, but the reference isn't. 

`references = _filter_definitions_by_reference()` is obvious that filter is in the same file, but `references = filter_definitions_by_reference()` can be a reference to a public function imported from another module.

python/src/aac/lang/definitions/source_location.py
```diff
@@ -0,0 +1,20 @@
+"""Parser Metadata class use to manage and identify the the source of definitions."""
+
+from attr import attrib, attrs, validators
+
+
+@attrs
+class SourceLocation:
+    """The position and ... of an AaC structure in the source.
```
Coffee2Bits, line 8:
```suggestion
    """The position and span of an AaC structure in the YAML source.
```

python/src/aac/plugins/gen_design_doc/gen_design_doc_impl.py
```diff
@@ -95,14 +91,12 @@ def _get_output_file_extension(template_filespec: str) -> str:
     return extension
 
 
-def _get_from_parsed_models(parsed_models: dict, aac_type: str) -> list:
-    aac_types = [util.get_models_by_type(m, aac_type) for m in parsed_models]
-    return list(flatten([m.values() for m in aac_types]))
+def _get_and_prepare_definitions_by_type(parsed_definitions: Definition, aac_type: str) -> list:
+    filtered_definitions = get_definitions_by_root_key(aac_type, parsed_definitions)
+    return list([definition.structure for definition in filtered_definitions])
```
Coffee2Bits, line 96: https://github.com/jondavid-black/AaC/pull/248#discussion_r840776253
```suggestion
    return [definition.structure for definition in filtered_definitions]
```

python/src/aac/plugins/gen_plant_uml/gen_plant_uml_impl.py
```diff
@@ -73,29 +73,33 @@ def puml_sequence(architecture_file: str, output_directory: Optional[str] = None
...
+        use_case_types = get_models_by_type(models, "usecase")
 
         properties = []
         participants = []
         sequences = []
 
         for use_case_title in use_case_types.keys():
             # declare participants
-            usecase_participants = util.search(use_case_types.get(use_case_title, {}), ["usecase", "participants"])
+            usecase_participants = search(use_case_types[use_case_title], ["usecase", "participants"])
```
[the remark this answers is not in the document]
Coffee2Bits:
```suggestion
            usecase_participants = search(use_case_types.get(use_case_title), ["usecase", "participants"])
```

python/src/aac/spec/core.py
```diff
@@ -96,10 +76,31 @@ def get_roots(reload: bool = False) -> list[str]:
         A list of strings, one entry for each root name in the AaC model specification.
     """
 
+    def get_field_name(fields_entry_dict: dict):
+        return fields_entry_dict.get("name")
+
```
Coffee2Bits, line 81:
```suggestion
```

python/src/aac/spec/core.py
```diff
@@ -96,10 +76,31 @@ def get_roots(reload: bool = False) -> list[str]:
...
     global ROOT_NAMES
 
     if len(ROOT_NAMES) == 0 or reload:
-        aac_data, _ = get_aac_spec()
-        ROOT_NAMES = search(aac_data["root"], ["data", "fields", "name"])
+        aac_definitions = get_aac_spec()
+        root_definition = get_definition_by_name("root", aac_definitions)
+        ROOT_NAMES = list(map(get_field_name, search_definition(root_definition, ["data", "fields"])))
```
Coffee2Bits, line 87:
```suggestion
        ROOT_NAMES = list(map(lambda field: field.name for field in fields, get_root_fields()))
```

python/src/aac/lang/_active_context.py
```diff
@@ -0,0 +1,165 @@
...
+    def get_root_fields(self) -> list[dict]:
+        """
+        Get the list of root keys as defined in the ActiveContext.
+
+        Returns a list of strings indicating the current root keys in the active context. These
+        keys may differ from those provided by the core spec since the ActiveContext applies definitions
+        from active plugins and user files, which may extend the set of root keys.
+
+        Returns:
+            A list of strings, one entry for each root name available in the ActiveContext.
```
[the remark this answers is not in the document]
Coffee2Bits: I'm good to add more extensive descriptions to return types. Looks like it's inline with the Google docstrings as seen in this example: https://google.github.io/styleguide/pyguide.html#doc-function-raises

python/src/aac/lang/language_context.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        Get the list of root keys as defined in the LanguageContext.
+
+        Returns a list of strings indicating the current root keys in the active context. These
+        keys may differ from those provided by the core spec since the LanguageContext applies definitions
+        from active plugins and user files, which may extend the set of root keys.
+        See :py:func:`aac.spec.get_root_keys()` for the list of root keys provided by the unaltered core AaC DSL.
+
+        Returns:
+            A list of strings, one entry for each root key in the LanguageContext.
```
Coffee2Bits, line 88:
```suggestion
        Returns a list of strings indicating the current root keys in the active context. 
        
        Returns:
            A list of strings, one entry for each root key in the LanguageContext.
            
            These keys may differ from those provided by the core spec since the LanguageContext applies definitions
            from active plugins and user files, which may extend the set of root keys.
            See :py:func:`aac.spec.get_root_keys()` for the list of root keys provided by the unaltered core AaC DSL.

```

python/src/aac/lang/language_context.py
```diff
@@ -0,0 +1,182 @@
...
+    def get_root_fields(self) -> list[dict]:
+        """
+        Get the list of root fields as defined in the LanguageContext.
+
+        Returns a list of dictionaries that consist of the root fields in the active context. These
+        keys may differ from those provided by the core spec since the LanguageContext applies definitions
+        from active plugins and user files, which may extend the set of root keys.
+
+        Returns:
+            A list of dictionaries, one dictionary for each root field including name and type.
```
Coffee2Bits, line 101:
```suggestion

        Returns:
            A list of dictionaries, one dictionary for each root field including name and type.
            
            These fields may differ from those provided by the core spec since the LanguageContext applies definitions
            from active plugins and user files, which may extend the set of root fields.
```

python/src/aac/lang/language_context.py
Coffee2Bits, line 129:
```suggestion
        Returns a list of strings indicating the currently defined primitive types in the active context.
        
        Returns:
            A list of strings, one entry for each primitive type defined in the LanguageContext.
            
            These types may differ from those provided by the core spec since the LanguageContext applies definitions
            from active plugins and user files, which may extend the set of root keys.
            See :py:func:`aac.spec.get_primitives()` for the list of root keys provided by the unaltered core AaC DSL.
```

python/src/aac/lang/language_context.py
```diff
@@ -0,0 +1,182 @@
...
+        Returns a list of strings of all definition types in the active context.
+
+        Returns:
+            A list of strings, one entry for each definition name available in the LanguageContext.
+        """
+        return list(map(lambda definition: definition.name, self.definitions))
+
+    def is_primitive_type(self, type: str) -> bool:
+        """Returns a boolean indicating if the type is defined as a primitive type."""
+        return remove_list_type_indicator(type) in self.get_primitive_types()
```
Coffee2Bits, line 146:
```suggestion
    def is_primitive_type(self, type: str) -> bool:
        """
        Returns a boolean indicating if the type is defined as a primitive type.
        
        Args:
            type (str): The type string.
            
        Returns:
            A boolean indicating if the string matches a primitive type defined in the context.
        """
        return remove_list_type_indicator(type) in self.get_primitive_types()
```

python/src/aac/lang/language_context.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def is_definition_type(self, type: str) -> bool:
+        """Returns a boolean indicating if the type is defined by another definition."""
+        return remove_list_type_indicator(type) in self.get_defined_types()
+
+    def get_definition_by_name(self, definition_name: str) -> Optional[Definition]:
+        """Return the definition corresponding to the argument, or None if not found."""
```
Coffee2Bits, line 153:
```suggestion
    def get_definition_by_name(self, definition_name: str) -> Optional[Definition]:
        """
        Return the definition corresponding to the argument, or None if not found.
        
        Args:
            definition_name (str): The definition name to search for.
            
        Returns:
            The definition corresponding to the name, or None if not found.
        """
```

python/src/aac/lang/language_context.py
```diff
@@ -0,0 +1,182 @@
...
+        if is_array_type(definition_name):
+            definition_name = remove_list_type_indicator(definition_name)
+
+        return self.definitions_name_dictionary.get(definition_name)
+
+    def _apply_extension_to_context(self, extension_definition: Definition) -> None:
+        """Apply the extension to the definitions it modifies in the LanguageContext."""
```
Coffee2Bits, line 160:
```suggestion
    def _apply_extension_to_context(self, extension_definition: Definition) -> None:
        """
        Apply the extension to the definitions it modifies in the LanguageContext.
        
        Args:
            The extension definition to apply to the context.
        """
```

python/src/aac/lang/language_context.py
```diff
@@ -0,0 +1,182 @@
...
+            self._apply_extension_to_context(definition)
+
+    def add_definitions_to_context(self, definitions: list[Definition]):
+        """
+        Add the list of Definitions to the list of definitions in the LanguageContext, any extensions are added last.
+
+        Args:
+            definitions: The list of Definitions to add to the context.
+        """
+
```
Coffee2Bits, line 67:
```suggestion
```

python/src/aac/plugins/gen_plugin/gen_plugin_impl.py
```diff
@@ -147,7 +149,7 @@ def _generate_template_files(plugin_type: str, template_properties: dict) -> dic
 
 
 def _prepare_and_generate_plugin_files(
-    parsed_models: dict[str, dict], plugin_type: str, architecture_file_path: str
+    definitions: list[Definition], plugin_type: str, architecture_file_path: str
 ) -> dict[str, list[TemplateOutputFile]]:
     """
     Parse the model and generate the plugin template accordingly.
```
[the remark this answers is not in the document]
Coffee2Bits: Pertaining to this comment https://github.com/jondavid-black/AaC/pull/248#discussion_r840779327

python/src/aac/plugins/validators/_validator_plugin.py
```diff
@@ -0,0 +1,23 @@
...
+        name: A string with the name of the command argument
+        definition: The AaC definition of the Validator Plugin
+        validation_function: The validation callback function
+        arguments: Any additional arguments to supply to the validation_function
+    """
+
+    name: str = attrib(validator=validators.instance_of(str))
+    definition: ParsedDefinition = attrib(validator=validators.instance_of(ParsedDefinition))
+    validation_function: callable = attrib(default=None, validator=validators.is_callable())
+    arguments: dict = attrib(default={}, validator=validators.instance_of(dict))
```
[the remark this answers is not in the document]
Coffee2Bits: I think we should remove arguments here, they are unique to the validation instance. For example, the same validation can have instances with different arguments like: `validation A` on `field Y` may have args `1` and `2`, but `validation A` on `field Z` may have args `3` nad `4`. The correct place to handle args is at the stage when applying the validator to the definition.

```suggestion
```

python/src/aac/plugins/validators/_validator_result.py
```diff
@@ -0,0 +1,18 @@
...
+
+    Attributes:
+        messages (list): A dictionary of definition names as keys to lists of messages
+        is_valid (bool): A bool indicating whether the definition is valid or not (True/False)
+    """
+
+    messages: list[str] = attrib(default=Factory(list), validator=validators.instance_of(list))
+    is_valid: bool = attrib(default=False, validator=validators.instance_of(bool))
+
+    def get_messages_as_string(self) -> str:
```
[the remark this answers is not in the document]
Coffee2Bits: Hmm. I'll think on this and circle back.

python/src/aac/plugins/validators/defined_references/__init__.py
```diff
@@ -0,0 +1,33 @@
+"""Validation plugin to ensure that each validation definition has a corresponding validation implementation."""
+
+from aac.package_resources import get_resource_file_contents
+from aac.plugins import hookimpl
+from aac.plugins.validators import ValidatorPlugin, get_validation_definition_from_plugin_definitions
+
+from ._validate_references import validate_references
```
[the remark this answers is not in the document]
Coffee2Bits: Good catch, it may be vestigial from some earlier import issues. Though, looking at the suggestion it may have been because of the length of the statement.

```suggestion
from aac.plugins.validators.defined_references._validate_references import validate_references
```

python/src/aac/plugins/validators/defined_references/defined_references.yaml
```diff
@@ -0,0 +1,33 @@
+validation:
+  name: Type references exist
+  description: Verifies that references to definitions and models are defined
```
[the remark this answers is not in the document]
Coffee2Bits: +1 will add with the batch suggestions.

python/src/aac/plugins/validators/required_fields/required_fields.yaml
```diff
@@ -0,0 +1,33 @@
+validation:
+  name: Required fields are present
+  description: Verifies every field declared in the `required` section are present and populated
+  behavior:
+    - name: Verify that definition fields marked as required are populated.
+      type: request-response
```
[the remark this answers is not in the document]
Coffee2Bits: Agreed, I put that more as a placeholder. We need to revise some of the core spec.

python/src/aac/plugins/validators/validator_implementation/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+@hookimpl
+def get_plugin_aac_definitions() -> str:
+    """
+    Return the plugins Aac definitions.
+
+    Returns:
+         string representing yaml extensions and data definitions employed by the plugin
+    """
+
```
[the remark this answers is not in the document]
Coffee2Bits: + 1 will add to batch suggestions

python/src/aac/spec/spec.yaml
```diff
@@ -49,6 +49,13 @@ data:
         the specification of any plugins you may have installed.
 
         You can extend any 'enum' and 'data' definition.
+    - name: validation
+      type: Validation
+      description: |
+        A reference to a validation plugin which will provide verification and validation
+        for the definition and any uses of it depending on which validation plugins are referenced.
```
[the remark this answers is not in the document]
Coffee2Bits: + 1 will add to batch suggestions

python/src/aac/spec/spec.yaml
```diff
@@ -124,6 +131,58 @@ data:
...
+  required:
+    - name
+---
+data:
+  name: ValidatorInput
+  fields:
+    - name: name
+      type: string
+      description: |
+        The name of the validation rule and plugin.
```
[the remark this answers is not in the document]
Coffee2Bits: +1 will add to batch suggestions

python/src/aac/spec/spec.yaml
```diff
@@ -124,6 +131,58 @@ data:
...
+---
+data:
+  name: ValidatorOutput
+  fields:
+    - name: name
+      type: string
+      description: |
+        The name of the validation rule and plugin.
```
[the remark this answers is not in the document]
Coffee2Bits: +1 will add to batch suggestions

python/src/aac/validate/_collect_validators.py
```diff
@@ -0,0 +1,68 @@
...
+from aac.lang.definitions.search import search_definition
+from aac.lang.definitions.structure import get_sub_definitions
+from aac.lang.hierarchy import get_definition_ancestry
+from aac.plugins.validators import ValidatorPlugin
+
+
+def get_applicable_validators_for_definition(
+    definition: Definition, validator_plugins: list[ValidatorPlugin], context: LanguageContext
+) -> list[ValidatorPlugin]:
+    """Traverse the definition and identify all applicable validator plugins."""
```
Coffee2Bits, line 16:
```suggestion
    """
    Return a list of all validator plugins that are applicable to the definition.
    
    Args:
        definition (Definition): The definition to search through.
        validator_plugins (list[ValidatorPlugin]): The list of available, registered validator plugins.
        context (LanguageContext): The language context
        
    Returns:
        A list of validator plugins that can be applied to the definition.
        
        The validator plugins are deemed applicable if the definition contains a fields or substructures that 
        are validated, such as field references.
    """
```

python/src/aac/parser/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""AaC Parser and related functions submodule."""
+
+from aac.parser._parse_source import parse, get_yaml_from_source
+from aac.parser._parser_error import ParserError
+
+__all__ = (
+    ParserError.__name__,
+    parse.__name__,
+    get_yaml_from_source.__name__,
+)
```
Coffee2Bits, line 10: 1. Update `parser` package to `parse`
2. Need to update parser error
  a. `aac.parser._parser_error` -> `aac.parser.parser_error`
3. Need to make the `parse_source` module public.

python/src/aac/plugins/validators/_common.py
```diff
@@ -0,0 +1,27 @@
...
+from aac.lang.definition_helpers import get_definitions_by_root_key
+from aac.plugins import PluginError
+
+
+def get_validation_definition_from_plugin_definitions(plugin_definitions_string: str) -> ParsedDefinition:
+    """
+    Parses the validation definition sourced from a validator plugin's definitions.
+
+    Args:
+        source_name (str): A name for the file source - necessary for error messages
```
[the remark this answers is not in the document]
Coffee2Bits: Good call.
```suggestion
```

python/tests/plugins/validators/test_validate_required_fields.py
```diff
@@ -0,0 +1,68 @@
...
+        validation_definitions = get_definitions_by_root_key("validation", parse(get_plugin_aac_definitions()))
+        self.assertEqual(1, len(validation_definitions))
+
+        expected_validator_plugin = ValidatorPlugin(
+            name="Required fields are present", definition=validation_definitions[0], validation_function=(lambda x: x)
+        )
+        self.assertEqual(expected_validator_plugin.name, actual_validator_plugin.name)
+        self.assertEqual(expected_validator_plugin.definition, actual_validator_plugin.definition)
+
+    def test_validate_validate_required_fields_no_missing_required_fields(self):
```
Coffee2Bits, line 36:
```suggestion
    def test_validate_required_fields_no_missing_required_fields(self):
```

python/src/aac/validate/_validation_result.py
```diff
@@ -0,0 +1,16 @@
...
+class ValidationResult:
+    """Represents the result of the validation of a single definition.
+
+    Attributes:
+        definitions (list): Definitions that were part of the validated content
+        messages (list): A dictionary of definition names as keys to lists of messages
+        is_valid (bool): A bool indicating whether the definition is valid or not (True/False)
+    """
+
+    definitions: list = attrib(validator=validators.instance_of(list))
```
cvchaparro, line 14: Should we do this?

```suggestion
    definitions: list = attrib(default=Factory(list), validator=validators.instance_of(list))
```

python/tests/lang/definitions/test_schema.py
```diff
@@ -0,0 +1,109 @@
+from unittest import TestCase
+
+from aac.lang.active_context_lifecycle_manager import get_active_context
+from aac.lang.definitions.schema import get_definition_schema, get_root_schema_definitions, get_schema_defined_fields
+from aac.spec import get_root_fields
+from tests.helpers.context import get_core_spec_context
+
```
cvchaparro, line 7: I'd like to group the `tests.helpers` imports together.

```suggestion

from tests.helpers.context import get_core_spec_context
```

python/tests/lang/test_context_manager.py
```diff
@@ -0,0 +1,23 @@
+from unittest import TestCase
+from aac.lang.language_context import LanguageContext
+
+from aac.lang.active_context_lifecycle_manager import get_active_context
+from aac.plugins.plugin_manager import get_plugin_definitions
+from aac.spec import get_aac_spec
+
+
+class TestContextManager(TestCase):
```
cvchaparro, line 9: I think we should rename this class and the file to reflect the updated filename in `src/`.

```suggestion
class TestActiveContextLifecycleManager(TestCase):
```

python/src/aac/validate/_validate.py
```diff
@@ -0,0 +1,100 @@
...
+from aac.parser import parse
+from aac.plugins.plugin_manager import get_validator_plugins
+from aac.plugins.validators import ValidatorPlugin, ValidatorResult
+from aac.validate._validation_error import ValidationError
+from aac.validate._validation_result import ValidationResult
+from aac.validate._collect_validators import get_applicable_validators_for_definition
+
+
+@contextmanager
+def validate_definitions(user_definitions: list[Definition]) -> ValidationResult:
```
cvchaparro, line 18: It looks like another comment that disappeared: Could we rename these context managers to something that reads better after `with` - i.e. `with validate_definitions` doesn't sound right, whereas `with validated_definitions` or something would sound more natural.

cvchaparro (changes requested):
Seems like some of my previous comments were removed - not sure how but I re-added some comments that were removed and the changes not applied, etc.

python/src/aac/cli.py
```diff
@@ -65,7 +66,7 @@ def _validate_cmd(model_file: str) -> PluginExecutionResult:
     """Run the built-in validate command."""
 
     def validate_model() -> str:
-        with validation(parser.parse, model_file):
+        with validate_source(model_file):
```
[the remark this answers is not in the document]
Coffee2Bits: I thought that I had already commented this, but it looks like the pattern is -ed if the context manager performs the action on enter or -ing if it performs the action on exit. Since the validation context executes the validation before entering the context scope, then we will want the -ed version. Thanks!
https://localcoder.org/naming-convention-for-context-manager-classes-with-blocks

python/src/aac/lang/definitions/definition.py
```diff
@@ -0,0 +1,84 @@
...
+        """Returns true if the definition is an enum definition."""
+        return self.get_root_key() == "enum"
+
+    def is_data(self) -> bool:
+        """Returns true if the definition is a data definition."""
+        return self.get_root_key() == "data"
+
+    def to_yaml(self) -> str:
+        """Return a yaml string based on the current state of the definition including extensions."""
+        return yaml.dump(self.structure, sort_keys=False)
```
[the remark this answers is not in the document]
Coffee2Bits: Yes, content refers to the source content, but it doesn't reflect changes to the structure. This to yaml function returns the yaml equivalent to the structure, even after its been modified by extensions and plugins.

python/src/aac/lang/references.py
```diff
@@ -0,0 +1,23 @@
...
+
+    Args:
+        source_definition (Definition): The definition that is being referenced
+        definitions_to_search (list[Definition]): The definitions to search through for references
+
+    Return:
+        A list of Definitions that reference the source_definition
+    """
+
+    def _filter_definitions_by_reference(definition_to_filter: Definition) -> bool:
```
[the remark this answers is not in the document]
Coffee2Bits: I wrote out my response in your other comment. Indentation is only applicable to the declaration of the inner function, not when it's called. However, the Python community seems to not care, and given our complexity rules then it should obvious enough that the function being referred to is an inner function 20 or so lines above and not a public module function. When in Rome...

python/src/aac/plugins/validators/_common.py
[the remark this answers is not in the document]
Coffee2Bits: Why would we need the source name here?

python/src/aac/plugins/validators/_validator_plugin.py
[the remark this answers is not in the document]
Coffee2Bits: It won't.

python/src/aac/validate/_validation_result.py
Coffee2Bits: Do we see a reason to make definitions optional for validation results?

python/src/aac/lang/definitions/structure.py
```diff
@@ -0,0 +1,110 @@
...
+                if not is_array_type(field_type):
+                    field_content = [field_content]
+                elif not type(field_content) is list:
+                    logging.error(
+                        f"Value '{field_content}' is not an array type like its defined type '{field_type}'. Bad value in the field '{field_name}' in definition '{source_definition.name}'."
+                    )
+                    continue
+
+                for field in field_content:
+                    _get_substructures(field_type_definition, field)
```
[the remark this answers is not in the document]
Coffee2Bits: You could determine that by keeping track of the definitions you've traversed so far since you always start from the top-level fields of a definition. I'd have to add some additional logic to support that -- why do you ask?

python/src/aac/plugins/validators/_validator_result.py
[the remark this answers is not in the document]
Coffee2Bits: I updated the plugin execution result to match, but I don't think it's worth forcing the classes to re-use some functions right now since the validator result will need to be adjusted for use with the LSP diagnostics feature.

python/src/aac/plugins/specifications/specifications.yaml
```diff
@@ -39,15 +39,15 @@ ext:
   dataExt:
     add:
       - name: requirements
-        type: RequirementReference[]
+        type: RequirementReference
```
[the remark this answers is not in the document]
Coffee2Bits: They look a leftover from a switch in the specifications definitions. Without the change, there is an array of `RequirementReference` wherein each element is an array of multiple requirement references per the `ids` field. So, there is an unnecessary extra dimension to the ids, which I removed with this change.

Here is the corresponding definition:
```yaml
data:
  name: RequirementReference
  fields:
    - name: ids
      type: string[]
  required:
    - ids
```

python/src/aac/lang/definitions/definition.py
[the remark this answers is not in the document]
cvchaparro: Ah, right, good point.

python/src/aac/validate/_validate.py
Coffee2Bits: Done

python/src/aac/lang/language_context.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+
+    def __attrs_post_init__(self):
+        """Post init hook for attrs classes."""
+        self.definitions_name_dictionary = {definition.name: definition for definition in self.definitions}
+
+    definitions: list[Definition] = attrib(default=Factory(list), validator=validators.instance_of(list))
+
+    # Private attribute - don't reference outside the class.
+    definitions_name_dictionary: dict[str, Definition] = attrib(
```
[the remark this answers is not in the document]
Coffee2Bits: https://www.attrs.org/en/stable/init.html#private-attributes
```
One thing people tend to find confusing is the treatment of private attributes that start with an underscore. attrs follows the doctrine that [there is no such thing as a private argument](https://github.com/hynek/characteristic/issues/6) and strips the underscores from the name when writing the __init__ method signature:
```

python/src/aac/lang/definitions/structure.py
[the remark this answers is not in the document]
cvchaparro: I was asking because I think that would make #252 a bit more straightforward. Essentially, I was curious if I'd be able to leverage this, for solving that.

python/src/aac/lang/language_context.py
```diff
@@ -0,0 +1,182 @@
...
+    def add_definition_to_context(self, definition: Definition):
+        """
+        Add the Definition to the list of definitions in the LanguageContext.
+
+        Args:
+            definition: The Definition to add to the context.
+        """
+        new_definition = copy.deepcopy(definition)
+
+        if new_definition.name not in self.definitions_name_dictionary:
```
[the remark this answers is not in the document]
Coffee2Bits: I think this is part of a larger discussion that we need to have regarding whether or not we should enforce unique definition names. For now, I'm ok leaving it as a debug log message.

python/src/aac/lang/_active_context.py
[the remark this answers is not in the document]
Coffee2Bits: My brain aggressively dumps memories, so I can't speak to a previous position without referring to a written conversation or decision. 

VSCode handles the back-tick formatting correctly from what I saw. The documentation I see indicates that they are necessary for some of the formatting and advanced docstring functionality in Sphinx, so I'm happy to argue for allowing backticks to support better documentation.

python/src/aac/lang/definitions/structure.py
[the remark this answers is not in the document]
Coffee2Bits: I think you'll need to create a separate function that effectively performs a look-up on a substructure by traversing the definition and identifying the target substructure. That traversal/lookup will look a lot like this function, but just different enough in scope and responsibilities to deserve its own public function.

I'm having trouble envisioning the problem without understanding what information VSCode will provide to that handler, but if we can access the entire definition then we can leverage keys like we do in search to narrow the traversal.

python/src/aac/lang/references.py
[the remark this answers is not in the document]
cvchaparro: 1. I don't think the underscore solves the issue of knowing it's an inner function because we also use a leading underscore for top-level functions, so we wouldn't have a way to distinguish that.
2. If we have an inner function separated from where it's called by 20 lines, then I'd argue that function is too big anyways. I realize that I have not commented on function length in the PR, mostly because we haven't established any guidelines for that but I agree with the idea of functions being very small as described in [this](https://vimeo.com/12643301) video (or, if you have the [Clean Code](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882/ref=sr_1_1?crid=B2J17SEP2OMS&keywords=clean+code&qid=1649196868&sprefix=clean+code%2Caps%2C125&sr=8-1) book, it's described in Chapter 3).

python/src/aac/validate/_validation_result.py
cvchaparro: Perhaps not. I forgot that adding `default` means definitions would be optional...

python/src/aac/lang/language_context.py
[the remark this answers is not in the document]
cvchaparro: Interesting. Thanks for sharing!

python/src/aac/lang/language_context.py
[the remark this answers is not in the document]
cvchaparro: Me too, I just thought it was something we should discuss at some point. I can convert this to a discussion. 👍

python/src/aac/lang/references.py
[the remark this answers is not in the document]
Coffee2Bits: My bad, I didn't make my last comment clear. I'll capitulate and not underscore my inner functions. 

For clarity, my position on the underscore is that it denotes that the function is defined in the current file regardless of whether it is defined at the top-level of the file/module or inside a function. In my missing comment, I made the point that the underscore before a function immediately denotes a non-public module function, which means that the function has to be defined in the current file which provides some immediate feedback to the developer. 

To your second point, you have to call the inner function after it's defined, so you're already a few lines from the declaration of the function name. Our flake complexity rules will handle this, so I'm not worried about it getting out of control.

python/tests/validate/test_validate.py
```diff
@@ -0,0 +1,77 @@
+from cgi import test
```
Coffee2Bits, line 1:
```suggestion
```

cvchaparro (approved):
Looks good

PR #278, opened 2022-03-29
Add command to print the active context

### LINKED ISSUE #250: Print Spec Should Include the Core Spec and the Active Context
The aac-core-spec command should be able to return the core specification as well as the active context, all active definitions and plugins applied to the core spec.

## AC:
- [x] aac-core-spec command is renamed print-core-spec
- [x] print-active-context command is added, and it returns the superset of
  - [x] The core specification modified by extensions
  - [x] Any definitions provided by active plugins

cvchaparro:
Closes #250

- Rename `aac-core-spec` command to `print-core-spec`.
- Add `print-active-context` command.


Coffee2Bits (approved):
Looks good!

PR #285, opened 2022-04-06
254 Validator Implementation Validation

### LINKED ISSUE #254: Create the Validator Implementation validator plugin
Create the "Validator Implementation" validator plugin. This validator plugin verifies that every validation definition has a corresponding python plugin implementation. (It validates validation plugins!)

## AC:
  - [x] `register_validators()`
  - [x]  validation success
  - [x]  validation failure
- [x] Validation fails if a validation definition doesn't have a corresponding, registered validation plugin

Coffee2Bits:
Closes #254

* Added a validator plugin that verifies validations have a corresponding implementation.
* Updated validation to also include all definitions in the context.

### DIFF
--- a/python/src/aac/plugins/validators/validator_implementation/__init__.py
+++ b/python/src/aac/plugins/validators/validator_implementation/__init__.py
@@ -17,7 +17,6 @@ def get_plugin_aac_definitions() -> str:
     Returns:
          string representing yaml extensions and data definitions employed by the plugin
     """
-
     return get_resource_file_contents(__package__, PLUGIN_YAML_FILE)
 
 
--- a/python/src/aac/plugins/validators/validator_implementation/__init__.py
+++ b/python/src/aac/plugins/validators/validator_implementation/__init__.py
@@ -4,6 +4,7 @@ from aac.package_resources import get_resource_file_contents
 from aac.plugins import hookimpl
 from aac.plugins.validators import get_validation_definition_from_plugin_definitions
 from aac.plugins.validators import ValidatorPlugin
+from aac.plugins.validators.validator_implementation._validator_implementation import validate_validator_implementations
 
 PLUGIN_YAML_FILE = "validator_implementation.yaml"
 
@@ -16,6 +17,7 @@ def get_plugin_aac_definitions() -> str:
     Returns:
          string representing yaml extensions and data definitions employed by the plugin
     """
+
     return get_resource_file_contents(__package__, PLUGIN_YAML_FILE)
 
 
@@ -28,4 +30,4 @@ def register_validators() -> ValidatorPlugin:
         A collection of data necessary to manage and execute validation plugins.
     """
     validation_definition = get_validation_definition_from_plugin_definitions(get_plugin_aac_definitions())
-    return ValidatorPlugin(validation_definition.name, validation_definition, (lambda x: x))
+    return ValidatorPlugin(validation_definition.name, validation_definition, validate_validator_implementations)
--- a/python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py
+++ b/python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py
@@ -7,7 +7,7 @@ from aac.plugins.plugin_manager import get_validator_plugins
 from aac.plugins.validators import ValidatorResult, ValidatorPlugin
 
 
-def validate_validator_implementations(definition_under_test: Definition, target_sub_definition: Definition, active_context: LanguageContext) -> ValidatorResult:
+def validate_validator_implementations(definition_under_test: Definition, target_sub_definition: Definition, active_context: LanguageContext, **validation_kwargs) -> ValidatorResult:
     """
     Validates that the validation definition has a corresponding registered ValidatorPlugin.
 
--- a/python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py
+++ b/python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py
@@ -42,7 +42,7 @@ def validate_validator_implementations(definition_under_test: Definition, target
             error_messages.append(registered_plugin_names)
             logging.debug(registered_plugin_names)
 
-    dicts_to_test = get_substructures_by_type(definition_under_test, target_sub_definition, active_context)
+    dicts_to_test = get_substructures_by_type(definition_under_test, target_schema_definition, active_context)
     list(map(validate_dict, dicts_to_test))
 
     return ValidatorResult(error_messages, len(error_messages) == 0)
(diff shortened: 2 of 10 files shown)


python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py
```diff
@@ -7,7 +7,7 @@
 from aac.plugins.validators import ValidatorResult, ValidatorPlugin
 
 
-def validate_validator_implementations(definition_under_test: Definition, target_sub_definition: Definition, active_context: LanguageContext) -> ValidatorResult:
+def validate_validator_implementations(definition_under_test: Definition, target_sub_definition: Definition, active_context: LanguageContext, **validation_kwargs) -> ValidatorResult:
```
Coffee2Bits, line 10: The validation_kwargs aren't used here, but I'm adding them for consistency amongst the validator functions. I don't know of a good way to enforce an interface for this function in python....

python/src/aac/plugins/validators/validator_implementation/__init__.py
```diff
@@ -16,6 +17,7 @@ def get_plugin_aac_definitions() -> str:
     Returns:
          string representing yaml extensions and data definitions employed by the plugin
     """
+
```
cvchaparro, line 20: For consistency, I think we should only have spaces after docstrings when the code that follows it is an inner function.

```suggestion
```

python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py
```diff
@@ -0,0 +1,48 @@
...
+from aac.plugins.validators import ValidatorResult, ValidatorPlugin
+
+
+def validate_validator_implementations(definition_under_test: Definition, target_sub_definition: Definition, active_context: LanguageContext, **validation_kwargs) -> ValidatorResult:
+    """
+    Validates that the validation definition has a corresponding registered ValidatorPlugin.
+
+    Args:
+        definition_under_test (Definition): The definition that's being validate. (Root validation definitions)
+        target_sub_definition (Definition): A definition with applicable validation. ("Validation" definition)
```
Coffee2Bits, line 16: Should we rename this argument for the validators?
```suggestion
        target_schema_definition (Definition): A definition with applicable validation. ("Validation" definition)
```

python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py
cvchaparro: What is the interface you're wanting to enforce?

python/src/aac/validate/_validate.py
```diff
@@ -56,7 +56,7 @@ def _validate_definitions(user_definitions: list[Definition]) -> ValidationResul
     def validate_each_definition(definition: Definition) -> list[ValidatorResult]:
         return _validate_definition(definition, registered_validators, active_context)
 
-    validator_results = list(flatten(map(validate_each_definition, user_definitions)))
+    validator_results = list(flatten(map(validate_each_definition, active_context.definitions)))
```
cvchaparro, line 59: I think I may be missing something here. Was this just a hole in a previous test that's being fixed?

python/tests/plugins/test_gen_json.py
```diff
@@ -10,6 +11,9 @@
 
 
 class TestGenJson(TestCase):
+    def setUp(self) -> None:
```
cvchaparro, line 14: If this `setUp` function is going to be needed in all our tests, we may benefit from having a `BaseTestCase` class that handles that so we don't forget. But I'm not sure if we actually need that in every test class.

cvchaparro (changes requested):
Looks good, overall. I just have a couple questions and a small change suggestion.

python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py
```diff
@@ -0,0 +1,48 @@
...
+from aac.plugins.plugin_manager import get_validator_plugins
+from aac.plugins.validators import ValidatorResult, ValidatorPlugin
+
+
+def validate_validator_implementations(definition_under_test: Definition, target_sub_definition: Definition, active_context: LanguageContext, **validation_kwargs) -> ValidatorResult:
+    """
+    Validates that the validation definition has a corresponding registered ValidatorPlugin.
+
+    Args:
+        definition_under_test (Definition): The definition that's being validate. (Root validation definitions)
```
cvchaparro, line 15:
```suggestion
        definition_under_test (Definition): The definition that's being validated. (Root validation definitions)
```

python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py
cvchaparro: I am in favor of this change.

python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py
Coffee2Bits: That the validator callback function, like the one on line 10, has the expected function arguments. Obviously, this is something that we'll ultimately catch in tests, but I would prefer something more in the developer's face like interpreter/compiler errors.

python/src/aac/plugins/validators/validator_implementation/__init__.py
Coffee2Bits: I don't immediately see black formatter or flake linter solutions that would support enforcing this. I'll try and keep the pattern, but it'll require manual policing.

python/tests/plugins/test_gen_json.py
Coffee2Bits: This is true only for testing plugins and more E2E scenarios. We can leverage creating our own LanguageContext for a number of the unit tests, so I'll make an issue to create a common test class for tests that involve the active context.

python/src/aac/validate/_validate.py
Coffee2Bits: So, I had a question that never got asked about whether or not we should reconsider how we do validation. The question was whether we should carry on the old pattern (that's what this change does) or use a more directed pattern like the one in main that only validates the target validations. 

Why does this matter? We'll have to consider the impact of validating every definition in the context when we consider what APIs we're leveraging in the LSP diagnostic feature -- which is driven by our validation system.

I haven't taken the time to really think about the problem so I'm reverting to old validator behavior until we elect to consciously change the user's view of validation behavior. That being said, we were still testing the core spec in the workflow since we run the validate command on the core spec.

cvchaparro: Can we move this to a discussion, then? I thought we were still using the older pattern but we should probably discuss the pros and cons of each approach to make sure we're not missing anything.

cvchaparro (approved):
Nice work. :smile:

PR #294, opened 2022-04-09
Bump the version number a minor increment

Coffee2Bits:
This pull request is to bump the version number a minor increment from 0.0.5 to 0.1.0.

* Validator overhaul minor version bump to 0.1.0

### DIFF
--- a/python/src/aac/__init__.py
+++ b/python/src/aac/__init__.py
@@ -1,7 +1,7 @@
 """The Architecture-as-Code tool."""
 import logging
 
-__version__ = "0.0.5"
+__version__ = "0.1.0"
 
 logging.basicConfig(
     format="%(asctime)s::%(module)s::%(filename)s::L%(lineno)d::%(levelname)s::%(message)s",
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -33,9 +33,10 @@
                     "scope": "machine-overridable",
                     "type": "string",
                     "enum": [
-                        "0.0.5"
+                        "0.0.5",
+                        "0.1.0"
                     ],
-                    "default": "0.0.5",
+                    "default": "0.1.0",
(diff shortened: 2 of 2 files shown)


python/src/aac/__init__.py
```diff
@@ -1,7 +1,7 @@
 """The Architecture-as-Code tool."""
 import logging
 
-__version__ = "0.0.5"
+__version__ = "0.1.0"
```
cvchaparro, line 4: Doing this will require updating the VSCode plugin [package.json](https://github.com/jondavid-black/AaC/blob/main/vscode_extension/package.json#L32-L40) file to allow `0.1.0` as a possible version.

Coffee2Bits: Why are we wanting to maintain that configuration item?

cvchaparro: Well, it gives users a way to switch between versions of AaC from within VSCode, if they want to, but given that user's can install different versions to do that the only other thing I can think of would be to ensure that a version of AaC that's compatible with the extension that is being used (i.e. pre-0.0.5 didn't have LSP functionality, so that wouldn't be a valid version and when we update the extension to use the `help-dump` command, versions prior to that wouldn't be valid). With that said, we can also remove it (and references to it) and just say the only supported version is the latest version of AaC. If you use a previous version, then some things may not work.

Coffee2Bits: > Well, it gives users a way to switch between versions of AaC from within VSCode

I believe that would require us to support managing aac versions through pip.

The only use for a version string that I can come up with is, like you pointed out, the minimum supported version. However, that's not really a user configuration item... I think the aac path is a sufficient tool for allowing users to choose their version of AaC. I've corrected the values for the test, but I'll also make an issue to remove the configuration item.

cvchaparro (approved):

PR #300, opened 2022-04-16
Update the VSCode extension to support connecting to LSP servers via TCP

### LINKED ISSUE #293: I Want the VSCode Extension to Support Connecting to LSP Servers via TCP
As a user of the AaC LSP and VSCode extension, I want the VSCode Extension to Support connecting to LSP servers via TCP, so that I can opt to use a local AaC LSP server or a remote AaC LSP server.

Food for thought: How can we get the extension to rety remote connections if they fail? The user can reload the window to effectively restart the extension, but we'll want to keep in mind the usability of that.

blocked by #227 

## AC:
- [x] The VSCode Extension has additional configuration settings
    - [x] LSP Server Mode
        - [x]  IO or TCP
    - [x] TCP Server Address
        - [x] Host and Port configuration options
        - [x] Host defaults to localhost value `127.0.0.1`
        - [x] Port defaults to `5007` 
- [x] If the extension is configured for IO mode, it runs the AaC LSP in IO mode
- [x] If the extension is configured for TCP mode, it connects to an LSP server via TCP
    - [x] If connection fails, an error message is reported to the user

cvchaparro:
Closes #293

Changes:
- Add a configuration item to allow selecting the type of LSP server to connect to.
- Add configuration items for TCP connection information.
- Update the VSCode extension to allow connecting to supported TCP servers.


python/src/aac/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 logging.basicConfig(
     format="%(asctime)s::%(module)s::%(filename)s::L%(lineno)d::%(levelname)s::%(message)s",
-    filename="aac.log",
+    # filename="aac.log",
```
cvchaparro, line 9:
```suggestion
    filename="aac.log",
```

python/src/aac/lang/lsp/server.py
```diff
@@ -25,12 +25,9 @@
...
 
 
-def start_lsp():
-    """Start the LSP server.
-
-    Args:
-        dev (bool): Whether to start the server and setup logging for development. (optional)
-    """
+# Temporary until issue #227 is complete
+def start_lsp(host=None, port=None):
```
cvchaparro, line 29:
```suggestion
def start_lsp():
```

python/src/aac/lang/lsp/server.py
```diff
@@ -39,7 +36,11 @@ def start_lsp():
 
     logging.info("Starting the AaC LSP server")
 
-    SERVER.start_io()
+    # Temporary until issue #227 is complete
+    if host and port:
+        SERVER.start_tcp(host, port)
+    else:
+        SERVER.start_io()
```
cvchaparro, line 43:
```suggestion
    SERVER.start_io()
```

python/src/aac/plugins/start_lsp/__init__.py
```diff
@@ -2,7 +2,7 @@
 # NOTE: It is safe to edit this file.
 # This file is only initially generated by aac gen-plugin, and it won't be overwritten if the file already exists.
 
-from aac.cli.aac_command import AacCommand
+from aac.cli.aac_command import AacCommand, AacCommandArgument
```
cvchaparro, line 5:
```suggestion
from aac.cli.aac_command import AacCommand
```

python/src/aac/plugins/start_lsp/__init__.py
```diff
@@ -22,6 +22,10 @@ def get_commands() -> list[AacCommand]:
             "start-lsp",
             "Start the AaC Language Server Protocol (LSP) server",
             start_lsp,
+            [
+                AacCommandArgument("--host", "The host to use for connections"),
+                AacCommandArgument("--port", "The port to use for connections"),
+            ]
```
cvchaparro, line 28:
```suggestion
```

python/src/aac/plugins/start_lsp/start_lsp_impl.py
```diff
@@ -8,8 +8,10 @@
 plugin_name = "start-lsp"
 
 
-def start_lsp() -> PluginExecutionResult:
+# Temporary until issue #227 is complete
+def start_lsp(host=None, port=None) -> PluginExecutionResult:
```
cvchaparro, line 12:
```suggestion
def start_lsp() -> PluginExecutionResult:
```

python/src/aac/plugins/start_lsp/start_lsp_impl.py
```diff
@@ -8,8 +8,10 @@
...
     """Start the IO LSP server."""
-    with plugin_result(plugin_name, lsp_server.start_lsp) as result:
+    # Temporary until issue #227 is complete
+    with plugin_result(plugin_name, lsp_server.start_lsp, host, port) as result:
```
cvchaparro, line 15:
```suggestion
    with plugin_result(plugin_name, lsp_server.start_lsp) as result:
```

cvchaparro (commented):
These changes were needed so I could test. Remove before merging.

vscode_extension/src/AacLanguageServer.ts
```diff
@@ -1,9 +1,17 @@
 import * as fs from "fs";
-import { ExtensionContext, workspace } from "vscode";
-import { LanguageClient, LanguageClientOptions, ServerOptions, Trace } from "vscode-languageclient/node";
+import * as net from "net";
+
+import { ExtensionContext, window, workspace } from "vscode";
+import { LanguageClient, LanguageClientOptions, ServerOptions, StreamInfo, Trace } from "vscode-languageclient/node";
 import { assertTrue, execShell, getConfigurationItem, showMessageOnError, getSemanticVersionNumber } from "./helpers";
 
+enum LspServerMode {
```
cvchaparro, line 8: I don't like this pattern but we can't get these values from the configuration information provided by the VSCode API (at least I haven't found a way to do it) so I created story #303 to give us more information about the configuration such as possible values when we have [enumerated values](https://github.com/jondavid-black/AaC/blob/cbb06b08525caa94af093263bdcb3fb52be5183a/vscode_extension/package.json#L52-L55) a user can pick.

vscode_extension/package.json
```diff
@@ -45,6 +45,31 @@
...
+                    "scope": "machine-overridable",
+                    "type": "string",
+                    "enum": [
+                        "IO",
+                        "TCP"
+                    ],
+                    "default": "IO",
+                    "description": "The method of communication with the LSP server."
+                },
+                "aac.lsp.tcp.host": {
```
Coffee2Bits, line 59: I'd expect host/port to be available for websockets too. Is there a reason to create per-protocol host/port config items or can these be generic amongst the protocols since any networking protocol will need host/port.
```suggestion
                "aac.lsp.host": {
```

vscode_extension/package.json
```diff
@@ -45,6 +45,31 @@
...
+                    "scope": "machine-overridable",
+                    "type": "string",
+                    "pattern": "^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$",
+                    "default": "127.0.0.1",
+                    "description": "The host to use for establishing a TCP connection with the LSP server."
+                },
+                "aac.lsp.tcp.port": {
```
Coffee2Bits, line 66: See above comment.
```suggestion
                "aac.lsp.port": {
```

Coffee2Bits (changes requested):
Looks good! Small question/discussion item to address.

vscode_extension/package.json
cvchaparro: I thought about this, too. Honestly, I was torn on whether it would matter that much but the main use case I thought of for having two different `host`/`port` configurations (one for TCP and one for WebSockets) was in case someone wanted to switch between LSP server implementations. In that case (with separate `host`/`port` configurations), they could just change the `serverMode` property to the style of server they want to use. Otherwise, they'd have a slightly more error-prone process of setting the `serverMode`, the `host`, and the `port` anytime they wanted to switch (this assumes they would not use the same `host`/`port` for both). It may be a small gain, but from a user's standpoint, I wouldn't want to necessarily have to update it, if I didn't have to.

Again, I doubt it would be a huge deal, I just thought it could offer a bit more ease of use, IMO.

Coffee2Bits (approved):

PR #301, opened 2022-04-16
Material Model Plug-In

jondavid-black:
The material model plug-in allows for modeling of the physical aspects of a system architecture and design.  It provides a rigorously configuration managed way to define parts that make up a system, while maintaining traceability to the underlying engineering.  That is to say, it seeks to be better than a huge spreadsheet that constitutes little more than a "list of stuff".  Standard AaC validation provides useful means of ensuring quality in the material model definition itself.

The original thought with this plugin was to provide a new command to generate a Bill of Materials (BOM), but after exploring the idea a bit it seemed better to separate the data modeling aspect from the reporting aspect.  After all, different teams and organizations may have different needs in the content and structure of their BOM.  It was decided that a better approach was to create a general purpose reporting mechanism that will allow teams to define their own BOM in a way that meets their needs.  A new discussion and PR will be created to capture this idea and the associated work.

- [x] Provide a README with a usage example.
- [x] Define parts, assemblies, and deployments.


python/model/material/kitchen.yaml
```diff
@@ -0,0 +1,34 @@
+import:
+  - ./apartment_provided.yaml
+  - ./amazon_parts.yaml
+
+assembly:
+  - name: Appliances
```
Coffee2Bits, line 6: The '-' here is causing the yaml to parse this value as a list as opposed to the expected key-value pair of the fields, `name`, `description`, and `parts`. The suggested fix allows AaC to successfully validate the `model/material/my_apartment.yaml` file for me.

I'll take the action to identify helpful error messages and cases for situations like this so that issues like this are more clear to the user.
```suggestion
  name: Appliances
```

Coffee2Bits (approved):
Looks good! I'm excited to see what we can can cook up with BOM reports.

PR #305, opened 2022-04-20
Root Key Validator

### LINKED ISSUE #258: Create the Root Key validation plugin
Create the "Root Key" validator plugin. This validator plugin verifies that every root key is defined in the "root" definition.

## AC:
  - [x] `register_validators()`
  - [x]  validation success
  - [x]  validation failure
- [x] Validation fails if a root key is used that isn't defined in the "root" definition
- [x] The validation is applied to the `data` AaC definition

Coffee2Bits:
Closes #258

* Added root key validator


python/src/aac/plugins/validators/root_keys/root_keys.yaml
```diff
@@ -0,0 +1,44 @@
+validation:
+  name: Root key is defined
+  description: Verify that the definition's root key is defined
+  behavior:
+    - name: Verify that the definition's root key is defined
+      type: request-response
+      description:
```
cvchaparro, line 7: If we're not adding a description, we should probably just remove this.

```suggestion
```

python/src/aac/plugins/validators/root_keys/root_keys.yaml
```diff
@@ -0,0 +1,44 @@
...
+        - scenario: Successfully validate definition root key
+          given:
+            - The ValidatorInput content consists of valid AaC definitions.
+            - The ValidatorInput contains some AaC definition that has a root key defined by the 'root' definition in the core specification.
+          when:
+            - The input is validation
+          then:
+            - The ValidatorOutput does not indicate any errors
+            - The ValidatorOutput does not indicate any warnings
+            - The ValidatorOutput indicates the validator plugin under test is valid
```
cvchaparro, line 24:
```suggestion
            - The ValidatorOutput indicates the definition under test is valid
```

python/src/aac/plugins/validators/root_keys/root_keys.yaml
```diff
@@ -0,0 +1,44 @@
...
+      output:
+        - name: results
+          type: ValidatorOutput
+      acceptance:
+        - scenario: Successfully validate definition root key
+          given:
+            - The ValidatorInput content consists of valid AaC definitions.
+            - The ValidatorInput contains some AaC definition that has a root key defined by the 'root' definition in the core specification.
+          when:
+            - The input is validation
```
cvchaparro, line 20:
```suggestion
            - The ValidatorInput is validated
```

python/src/aac/plugins/validators/root_keys/root_keys.yaml
```diff
@@ -0,0 +1,44 @@
...
+            - The ValidatorOutput does not indicate any errors
+            - The ValidatorOutput does not indicate any warnings
+            - The ValidatorOutput indicates the validator plugin under test is valid
+        - scenario: Successfully validate an extended root key
+          given:
+            - The ValidatorInput content consists of valid AaC definitions.
+            - The ValidatorInput content of an extension definition that adds a new root key.
+            - The ValidatorInput contains some AaC definition that has a root key that has been added to the 'root' definition in the core specification.
+          when:
+            - The input is validation
```
cvchaparro, line 31:
```suggestion
            - The ValidatorInput is validated
```

python/src/aac/plugins/validators/root_keys/root_keys.yaml
```diff
@@ -0,0 +1,44 @@
...
+          then:
+            - The ValidatorOutput does not indicate any errors
+            - The ValidatorOutput does not indicate any warnings
+            - The ValidatorOutput indicates the validator plugin under test is valid
```
cvchaparro, line 35:
```suggestion
            - The ValidatorOutput indicates the definition under test is valid
```

python/src/aac/plugins/validators/root_keys/root_keys.yaml
```diff
@@ -0,0 +1,44 @@
...
+        - scenario: Fail to validate definition references
```
cvchaparro, line 36:
```suggestion
        - scenario: Fail to validate definitions that reference keys that are not defined by the 'root' definition
```

python/src/aac/plugins/validators/root_keys/root_keys.yaml
```diff
@@ -0,0 +1,44 @@
...
+          given:
+            - The ValidatorInput content consists of otherwise valid AaC definitions.
+            - The ValidatorInput contains at least one AaC definition that has a root key not defined by the 'root' definition in the core specification.
+          when:
+            - The ValidatorInput is validated
+          then:
+            - The ValidatorOutput has errors
+            - The ValidatorOutput errors indicate that there the root key is not defined in the root keys definition.
```
cvchaparro, line 44:
```suggestion
            - The ValidatorOutput errors indicate that the root key is not defined in the root keys definition.

```

python/src/aac/plugins/validators/root_keys/_validate_root_keys.py
```diff
@@ -0,0 +1,32 @@
...
+        target_schema_definition (Definition): A definition with applicable validation.
+        active_context (LanguageContext): The active context.
+
+    Returns:
+        A ValidatorResult containing any applicable error messages.
+    """
+    error_messages = []
+    active_context_keys = get_active_context().get_root_keys()
+
+    # We're expecting only full definition dicts -- so the root key is the first value
```
cvchaparro, line 24: We should encode this expectation in a test rather than a comment.

python/src/aac/plugins/validators/root_keys/_validate_root_keys.py
```diff
@@ -0,0 +1,32 @@
...
+    Args:
+        definition_under_test (Definition): The definition that's being validated.
+        target_schema_definition (Definition): A definition with applicable validation.
+        active_context (LanguageContext): The active context.
+
+    Returns:
+        A ValidatorResult containing any applicable error messages.
+    """
+    error_messages = []
+    active_context_keys = get_active_context().get_root_keys()
```
cvchaparro, line 22:
```suggestion
    active_context_root_keys = get_active_context().get_root_keys()
```

python/src/aac/plugins/validators/root_keys/_validate_root_keys.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # We're expecting only full definition dicts -- so the root key is the first value
+    root_key = definition_under_test.get_root_key()
+
+    if root_key not in active_context_keys:
```
cvchaparro, line 27:
```suggestion
    if root_key not in active_context_root_keys:
```

python/src/aac/plugins/validators/root_keys/_validate_root_keys.py
```diff
@@ -0,0 +1,32 @@
+import logging
+
+from aac.lang.language_context import LanguageContext
+from aac.lang.active_context_lifecycle_manager import get_active_context
+from aac.lang.definitions.definition import Definition
+from aac.plugins.validators import ValidatorResult
+
+
+def validate_root_keys(definition_under_test: Definition, target_schema_definition: Definition, active_context: LanguageContext, *validation_args) -> ValidatorResult:
```
cvchaparro, line 9: I've noticed a few places where we alternate between using the `active_context` parameter and calling `get_active_context()`. I expect this is effectively the same in our tests because we're always resetting the general active context but I wonder if it makes sense for us to be passing `active_context` if we're going to default to getting the active context via the `get_active_context()` function?

Having both options seems error-prone.

python/src/aac/plugins/validators/root_keys/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+
+@hookimpl
+def get_plugin_aac_definitions() -> str:
+    """
+    Return the plugins Aac definitions.
+
+    Returns:
+         string representing yaml extensions and data definitions employed by the plugin
+    """
+
```
cvchaparro, line 20:
```suggestion
```

cvchaparro (changes requested):

python/src/aac/lang/definitions/definition.py
```diff
@@ -75,10 +75,6 @@ def is_enum(self) -> bool:
         """Returns true if the definition is an enum definition."""
         return self.get_root_key() == "enum"
 
-    def is_data(self) -> bool:
```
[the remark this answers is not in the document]
Coffee2Bits: It's unused at the moment.

python/src/aac/plugins/validators/root_keys/_validate_root_keys.py
Coffee2Bits: Good call out. I think the fix is to rename arguments from `active_context` to `language_context`, which should leave `get_active_context()` as the only reference to the Active Context. 

This change will need to be done across all of the validator plugins -- we should coordinate before I make that change in the PR, or we can make a small issue for it.

python/src/aac/plugins/validators/root_keys/_validate_root_keys.py
Coffee2Bits: This comment is old. 🙃
```suggestion
```

python/src/aac/plugins/validators/root_keys/_validate_root_keys.py
cvchaparro: I think creating a small dedicated issue for it would be the best approach, personally.

cvchaparro (approved):
Nice work!

PR #312, opened 2022-04-21
Update the VSCode extension to leverage the help-dump command

### LINKED ISSUE #289: Update the VSCode Extension to leverage help-dump for AaC commands
Update the VSCode Extension to leverage help-dump for AaC commands to reduce the number of calls to the AaC package and to provide better type support for command arguments.

## AC:
- [x] The VSCode extension leverages the `aac help-dump` to parse for available commands and arguments.

cvchaparro:
Closes #289

Changes:
- Get the AaC commands, arguments, and descriptions from the output of the `help-dump` command.
- Specify an absolute path for the log filename.

### DIFF
--- a/python/src/aac/__init__.py
+++ b/python/src/aac/__init__.py
@@ -6,7 +6,6 @@ __version__ = "0.1.0"
 
 logging.basicConfig(
     format="%(asctime)s::%(module)s::%(filename)s::L%(lineno)d::%(levelname)s::%(message)s",
-    filename="aac.log",
     encoding="utf-8",
     level=logging.DEBUG,
     datefmt="%m/%d/%Y %H:%M:%S",
--- a/python/src/aac/__init__.py
+++ b/python/src/aac/__init__.py
@@ -1,11 +1,13 @@
 """The Architecture-as-Code tool."""
 
 import logging
+import os
 
 __version__ = "0.1.0"
 
 logging.basicConfig(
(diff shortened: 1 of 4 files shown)


python/src/aac/__init__.py
```diff
@@ -1,12 +1,13 @@
...
 
 import logging
+import os
 
 __version__ = "0.1.0"
 
 logging.basicConfig(
     format="%(asctime)s::%(module)s::%(filename)s::L%(lineno)d::%(levelname)s::%(message)s",
-    filename="aac.log",
+    filename=os.path.join(os.path.dirname(__file__), "aac.log"),
```
cvchaparro, line 10: Make sure this works even when the `aac` tool is installed globally.

Coffee2Bits (changes requested):
Looks good! Only issue looks like my fault. The VSCode extension is still trying to launch the lsp with `start-lsp` which doesn't work since we're expecting `start-lsp-io` now. I'm not sure why that change isn't there, but you'll need to change 
`start-lsp` to `start-lsp-io` here
https://github.com/jondavid-black/AaC/blob/f3e2d2d77bb6a0dec4f1119d6445d7396fef24b2/vscode_extension/src/AacLanguageServer.ts#L109

Coffee2Bits (approved):
We spoke about the lsp fix being present in your other PR. 
You're good to go! 👍

PR #316, opened 2022-04-29
Update Validator Plugin Function Signature

### LINKED ISSUE #314: Update the argument name active_context in the validator plugin callback
Update the argument name active_context in the validator plugin callback.

> Good call out. I think the fix is to rename arguments from `active_context` to `language_context`, which should leave `get_active_context()` as the only reference to the Active Context. 

_Originally posted by @Coffee2Bits in https://github.com/jondavid-black/AaC/pull/305#discussion_r857132302_

Examples of the function signatures to update:
  https://github.com/jondavid-black/AaC/blob/b5ced0eaaac8f60039252c6286f7d9397b329202/python/src/aac/plugins/validators/validator_implementation/_validator_implementation.py#L10
  
 https://github.com/jondavid-black/AaC/blob/b5ced0eaaac8f60039252c6286f7d9397b329202/python/src/aac/plugins/validators/required_fields/_validate_required_fields.py#L9

## AC:
- [x] Every validator plugin validation callback function argument `active_context` is changed to `language_context`
- [x] Validation plugin callback function docstrings are correspondingly updated.

Coffee2Bits:
Closes #314 
* Update Validator Plugin Function Signature
* Update AaC version number

### DIFF
--- a/python/src/aac/__init__.py
+++ b/python/src/aac/__init__.py
@@ -3,7 +3,7 @@
 import logging
 import os
 
-__version__ = "0.1.0"
+__version__ = "0.1.1"
 
 logging.basicConfig(
     format="%(asctime)s::%(module)s::%(filename)s::L%(lineno)d::%(levelname)s::%(message)s",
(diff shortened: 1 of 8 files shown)


jondavid-black (approved):
Simple updates.  Looks good to me.

cvchaparro (approved):
Changes look good and validation still works as expected.

PR #330, opened 2022-05-10
Rename data to schema

Coffee2Bits:
* Renamed 'data' definitions to 'schema'
* Renamed 'dataExt' definitions to 'schemaExt'
* Updated several plugin templates to incorporate the schema -> data change

### DIFF
--- a/python/tests/lang/definitions/test_structure.py
+++ b/python/tests/lang/definitions/test_structure.py
@@ -67,7 +67,7 @@ class TestDefinitionStructures(TestCase):
     def test_get_definition_schema_components_with_data(self):
         test_context = get_core_spec_context()
 
-        data_definition = test_context.get_definition_by_name("data")
+        data_definition = test_context.get_definition_by_name("schema")
 
         # Per the core spec, we'd expect Field and ValidationReference
         field_definition = test_context.get_definition_by_name("Field")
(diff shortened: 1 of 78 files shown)


Coffee2Bits:
Everyone was onboard with renaming `data` to `schema` so I went ahead and give it a shot to determine if there were any unexpected barriers. Turns out it was pretty easy, so here is the PR with the name switch implemented.

jondavid-black (approved):
Wow, that was a lot of updates.  Everything looks good to me.  I think this change makes our data modeling more clear to the user...thanks for bringing up the idea.

Coffee2Bits:
Pushed an additional set of changes based on Cameron's feedback.

python/tests/lang/definitions/test_structure.py
```diff
@@ -20,14 +20,14 @@ def test_get_substructures_by_type_with_user_defined_data(self):
 
         test_string_field_entry = create_field_entry("TestStringField", "string")
         test_file_field_entry = create_field_entry("TestFileField", "file")
-        test_data_definition = create_data_definition("TestData", fields=[test_string_field_entry, test_file_field_entry])
+        test_schema_definition = create_schema_definition("TestData", fields=[test_string_field_entry, test_file_field_entry])
```
cvchaparro, line 23: I won't nitpick this, right now, but this name - i.e. `TestData` - reveals a possible area of improvement. Namely, when naming test definitions (or whatever else), it might be better to name them something generic enough that a change like this - i.e. `data` -> `schema` - wouldn't mean we have to go through each name and update it.

Coffee2Bits: I agree. This PR has had me stewing on some ways to combat all the strings.

PR #336, opened 2022-05-16
Correctly handle definition sources

### LINKED ISSUE #322: Update the Definition class with source URI attribute.
Definitions need a clearer link to source documents. Currently, definitions are only linked to source documents via lexemes, which provides an awkward interface for determining the source URI for a definition.

## AC:
- [x] The `Definition` class has a new attribute that contains the URI path to the definition's source
- [x] The language context's `get_definitions_by_file_uri` method is updated to account for the new attribute
- [x] `parse` is updated to populate the new attribute when parsing files
  - [x] coordinate updating this #323

### LINKED ISSUE #323: Parse incorrectly attributes the source document
When `parse` is called on an AaC file. This parsing method also incorrectly attributes source documents to imported definitions, for instance definitions defined in file B, but imported by file A will be attributed as being sourced by file B.

## AC:
- [x] `parse` correctly attributes the source URI to imported definitions.

cvchaparro:
Closes #322
Closes #323

Changes:
- Add a `source_uri` attribute to `Definition`s so we can get definitions from specific file URIs
- Update the parser to assign the correct source in which a `Definition` was defined


Coffee2Bits (approved):
Looks good! Good job!

PR #340, opened 2022-05-19
Addressed a small error case in subcomponent validator

Coffee2Bits:
Found an issue where models without the field 'type' defined would cause the subcomponent validator to exception-out and didn't provide helpful feed back on the issue, since the required fields validator plugin wasn't allowed to execute/return the error about the missing field.

This is a symptom of a larger issue that I have documented in #341.

### DIFF
--- a/python/src/aac/__init__.py
+++ b/python/src/aac/__init__.py
@@ -3,7 +3,7 @@
 import logging
 import os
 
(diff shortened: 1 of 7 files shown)


cvchaparro (approved):
Nice work!

PR #346, opened 2022-05-31
Overhaul the template engine

### LINKED ISSUE #319: Overhaul the template engine
The template paradigm we currently have is outdated and restrictive. The templates package should encompass the entirety of the template functionality, and plugins should be able to provide their own accompanying templates.

## AC:
- [x] Plugins can load templates from their own package
- [x] First-party plugin templates are moved to their respective plugins' packages
- [x] `TemplateOutputFile`s attribute `parent_dir` is changed to `output_directory` and is required.
- [x] `write_generated_templates_to_file` uses the `output_directory` attribute for the filepath
- [x] `load_default_templates` is removed.
- [x] All template-focused python modules and functionality is contained in the `aac.templates` package.

cvchaparro:
Closes #319

Changes:
- Rename `parent_dir` attribute in `TemplateOutputFile` to `output_directory` and make it required.
- Move plugin templates to the corresponding plugin package.
- Move `template_engine` module to the `templates` package and rename to `engine` module.
- Various changes to `aac.templates.engine` module.
- Minor clean up


python/src/aac/plugins/gen_protobuf/gen_protobuf_impl.py
```diff
@@ -53,7 +50,7 @@ def generate_protobuf_messages(architecture_file: str, output_directory: str) ->
 
     # Validate the source AaC file and its contents
     with validated_source(architecture_file) as validation_result:
-        loaded_templates = load_default_templates("gen_protobuf")
+        loaded_templates = load_templates(__package__, ".")
```
Coffee2Bits, line 53: What's the motivation to not giving Gen-Protobuf a templates directory? Is it because it only has the one template?

python/tox.ini
```diff
@@ -17,7 +17,7 @@ per-file-ignores =
     # Ignore docstring errors in tests
     *tests/*:D100,D101,D102,D104,D200,ANN101,ANN201,W293,W291,ANN001,E731
     *setup.py:D100
-    *src/aac/templates/*/__init__.py:D104
+    *src/aac/plugins/**/templates/**:D104
```
Coffee2Bits, line 20: This line may not be adequate if we don't maintain the pattern, like with GenProtobuf.

Issue #351, opened 2022-06-01
I want the REST API to expose all AaC commands


Coffee2Bits:
I want the REST API to expose all AaC commands.

Blocked by #344 

## AC:
- [x] The REST API exposes all AaC Commands
  - [x] Any command that is executed via the REST API has its output returned in the response
  - [x] Responses leverage HTTP status codes to indicate success, error, or other issues
    - [x] 200 OK on Success
    - [ ] 400 Bad Request
    - [x] 404 Not Found 
    - [x] 500 Internal Server Error
- [x] There are functional tests that exercise the AaC Commands via the REST API

PR #354, opened 2022-06-02
REST API Plugin

### LINKED ISSUE #344: I want a REST API plugin for the AaC Python package
I want a REST API plugin for the AaC Python package, so that the AaC package can be used with web-based frontends.

## AC:
- [x] There is a new REST API plugin
- [x] The REST API plugin has a command that starts up a RESTful service
  - [x] The command has an optional host argument 
  - [x] The command has an optional port argument
- [x] The REST API exposes a read-only list of AaC files in the context
- [x] There are functional tests that demonstrate the correctness of the REST API

Coffee2Bits:
Closes #344 
* Added a REST API plugin
  * Has an optional --host argument
  * Has an optional --port argument
* Added a basic, first route that lists all of the files for definitions in the active context
* Adjusted the parse function to allow overriding the source_uri of definitions
* Added a test for the files endpoint


python/src/aac/parser/_parse_source.py
```diff
@@ -8,18 +8,20 @@
...
 from aac.lang.definitions.source_location import SourceLocation
 
 
-def parse(source: str) -> list[Definition]:
+def parse(source: str, source_uri: Optional[str] = None) -> list[Definition]:
     """Parse the Architecture-as-Code (AaC) definition(s) from the provided source.
 
     Args:
         source (str): Must be either a file path to an AaC yaml file or a string containing AaC definitions.
+        source_uri (Optional[str]): Overrides and sets the source_uri for AaC sources that are passed in as strings.
```
Coffee2Bits, line 24:
```suggestion
        source_uri (Optional[str]): Overrides and sets the source_uri
```

Coffee2Bits:
I've had to add some duct tape to address shortcomings how plugins provide info about themselves and how the parser is structured. I'll create a story to address those shortcomings before putting this PR into review.

python/src/aac/plugins/plugin_manager.py
```diff
@@ -91,7 +93,9 @@ def get_plugin_definitions() -> list[Definition]:
     plugin_definitions = []
     for plugin_definition in plugin_definitions_as_yaml:
         if len(plugin_definition) > 0:
-            plugin_definitions.append(parser.parse(plugin_definition))
+            # Temporary setting until plugins can be updated to provide better contextual information
```
cvchaparro, line 96: We should update the plugin change story's AC to require removing these comments.

python/src/aac/plugins/rest_api/rest_api.yaml
```diff
@@ -0,0 +1,31 @@
...
+      description: Start a RESTful interface for interacting with and managing AaC.
+      input:
+        - name: --host
+          type: string
+          python_type: str
+          description: Set the hostname of the service. Useful for operating behind proxies.
+        - name: --port
+          type: int
+          python_type: int
+          description: The port to bing the RESTful service to.
```
cvchaparro, line 16:
```suggestion
          description: The port to which the RESTful service will be bound.
```

python/src/aac/plugins/rest_api/aac_rest_api_impl.py
```diff
@@ -0,0 +1,29 @@
...
+        port (Optional[int]): The port to bing the RESTful service to.
+    """
+    with plugin_result(plugin_name, _start_restful_service, host, port) as result:
+        return result
+
+
+def _start_restful_service(host: Optional[str], port: Optional[int]) -> None:
+    """Start the RESTful interface service."""
+    uvicorn.run(app, host=host or "0.0.0.0", port=port or 8000)
+    return "Successfully started and stopped the RESTful API."
```
cvchaparro, line 29: Since a user would only see this after the REST API was terminated, I think it would be better to say something like:

```suggestion
    return "REST API connection closed."
```

python/src/aac/plugins/rest_api/aac_rest_api_impl.py
```diff
@@ -0,0 +1,29 @@
...
+plugin_name = "aac-rest-api"
+
+
+def rest_api(host: Optional[str], port: Optional[int]) -> PluginExecutionResult:
+    """
+    Start a RESTful interface for interacting with and managing AaC.
+
+    Args:
+        host (Optional[str]): Set the hostname of the service. Useful for operating behind proxies.
+        port (Optional[int]): The port to bing the RESTful service to.
```
cvchaparro, line 20:
```suggestion
        port (Optional[int]): The port to which the RESTful service will be bound.
```

cvchaparro (changes requested):

python/src/aac/plugins/rest_api/aac_rest_api_impl.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    with plugin_result(plugin_name, _start_restful_service, host, port) as result:
+        return result
+
+
+def _start_restful_service(host: Optional[str], port: Optional[int]) -> None:
+    """Start the RESTful interface service."""
+    uvicorn.run(app, host=host or "0.0.0.0", port=port or 8000)
```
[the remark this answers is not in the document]
Coffee2Bits: Correct, uvicorn does block until killed. It's the best case since it allows the active context to persist between REST calls -- something that's lacking from the CLI. 

We don't want to link the LSP and the REST API -- the implementation would be too difficult and complex, but if we let the REST server alter filesystem files, then the LSP editor should pick up the changes and effectively synchronize to the new content. The only downside for letting the REST server stomp files is that it will cause issues with the VSCode editor and unsaved changes, which the user will ultimately have learn to deal with. Think of the REST API as an alternative to the LSP -- the user will have to use one or the other, but they can switch between modes.

python/src/aac/plugins/plugin_manager.py
Coffee2Bits: Done

python/src/aac/plugins/validators/_common.py
```diff
@@ -11,12 +11,12 @@ def get_validation_definition_from_plugin_definitions(plugin_definitions_string:
     Parses the validation definition sourced from a validator plugin's definitions.
 
     Args:
-        source_name (str): A name for the file source - necessary for error messages
         plugin_definitions_string (str): The definitions as a yaml string
 
     Returns:
         The validation definition for the plugin.
     """
+    # Temporary fix until plugins can provide better contextual information.
```
Coffee2Bits, line 19:
```suggestion
```

python/src/aac/plugins/rest_api/aac_rest_api_impl.py
Coffee2Bits: I'm going to disagree with this suggestion. I feel that the original better fits the pattern of other command success messages and is more descriptive as to what actually occurred. Users will be presented with realtime connection status information from the server's output to the shell.

cvchaparro: I don't disagree with your claim that it follows the pattern in other plugins but for this plugin, I think it's different given the long-running nature of it. Specifically, all of our other plugins have output that will be displayed almost immediately. In the case of the REST API, I don't see what value I'd get as a user to get a message indicating that the server was started and stopped when I saw that via other means - i.e. interacting with the server, the server's output, etc.

cvchaparro (changes requested):

python/src/aac/plugins/rest_api/aac_rest_api_impl.py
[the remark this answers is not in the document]
cvchaparro: For the first item, we decided on updating the VSCode extension to start the REST server after which we will start the LSP server via a call on the REST API (which will require completion of #351).

cvchaparro (approved):
Nice work!

PR #356, opened 2022-06-02
Update required fields to be specified as arguments to the required fields validator

### LINKED ISSUE #334: Updated required fields to be specified as arguments to the required fields validator
Updated required fields to be specified as arguments to the required fields validator, so that users don't have duplicate defining required fields.

## AC:
- [ ] The field `required` is removed from all definitions
- [x] The required fields validator plugin takes field names as arguments
- [ ] Existing definitions in the repository are updated

cvchaparro:
Closes #334

Changes:
- Remove the `required` field on `schema` definitions.
- Update access location for required fields from the `required` field defined on `schema`s to getting them from the validation arguments.
- Move getting required fields for a definition to the `required_fields` validation plugin.
- Update documentation to account for changes in the handling of required fields.

### DIFF
--- a/python/tests/plugins/validators/test_required_fields.py
+++ b/python/tests/plugins/validators/test_required_fields.py
@@ -84,7 +84,7 @@ class TestRequiredFieldsPlugin(TestCase):
 
         assert_validator_result_failure(actual_result, "fields", "not populated")
 
-    def test_validate_required_fields_added_by_extension(self):
+    def test_required_fields_added_by_extension(self):
         test_active_context = get_active_context()
 
         schema_field_name = "TestField1"
@@ -94,12 +94,7 @@ class TestRequiredFieldsPlugin(TestCase):
         test_active_context.add_definition_to_context(test_definition)
 
         required_fields_definition = test_active_context.get_definition_by_name(test_definition.name)
-        required_fields = get_required_fields(required_fields_definition)
-
-        self.assertEqual(len(required_fields), 0)
-
-        actual_result = validate_required_fields(test_definition, required_fields_definition, test_active_context, *required_fields)
-        assert_validator_result_success(actual_result)
+        self.assertEqual(len(get_required_fields(required_fields_definition)), 0)
 
         schema_extension_field_name = "TestField2"
         schema_extension_field = create_field_entry(schema_extension_field_name, "string")
@@ -112,9 +107,6 @@ class TestRequiredFieldsPlugin(TestCase):
 
         self.assertEqual(len(required_fields), 1)
         self.assertEqual(required_fields[0], schema_extension_field_name)
-
-        actual_result = validate_required_fields(test_definition, required_fields_definition, test_active_context, *required_fields)
-        assert_validator_result_success(actual_result)
 
     def test_is_field_populated(self):
 


cvchaparro:
There are several places where the required fields validator name is hard-coded - fix this by (ideally) making the validator name importable.

python/src/aac/lang/definitions/extensions.py
```diff
@@ -106,18 +106,34 @@ def _remove_schema_fields(target_fields: list, fields_to_remove: list):
 def _add_extension_required_fields_to_defintion(target_definition_fields: dict, extension_additional_content_fields: dict) -> None:
     """Add any additional required fields from the extension to the target definition."""
     extension_required_fields = extension_additional_content_fields.get("required")
+    definition_validations = target_definition_fields.get("validation") or []
+
+    if not definition_validations:
+        target_definition_fields["validation"] = []
+
+    from aac.plugins.validators.required_fields import REQUIRED_FIELDS_VALIDATION_STRING
```
cvchaparro, line 114: I hate doing this but I haven't found another way to resolve the circular import issue. I think we're going to need to put some extra thought into how we handle required fields beyond what we originally planned to do.

Coffee2Bits: I think this is an ok solution, it's not really any worse than the earlier idea. However, I'd make it a private function rather than writing it inline twice.

python/src/aac/validate/_validate.py
```diff
@@ -94,12 +95,14 @@ def _validate_definition(
 
 
 def _apply_validator(
-    definition_structure_dict: dict, target_schema_definition: Definition, context: LanguageContext, validator_plugin: ValidatorPlugin
```
Coffee2Bits, line 97: Good catch!

python/src/aac/plugins/gen_design_doc/gen_design_doc_impl.py
```diff
@@ -9,6 +9,7 @@
 from aac.lang.definition_helpers import get_definitions_by_root_key
 from aac.lang.definitions.definition import Definition
 from aac.plugins.plugin_execution import PluginExecutionResult, plugin_result
+from aac.plugins.validators.required_fields import get_required_fields
```
Coffee2Bits, line 12: Let's move the `get_required_fields` function to somewhere in the `aac.lang.definitions` package if we're going to be using it outside of the validator plugin.

I've been attaching little helpful methods like this to the Definition class, which is in part due to my OOP bias. However, I'd be ok with changing the overall style of those methods to be more functional, I just want the helper functions colocated since they're really similar in scope and goal.

python/tests/lang/test_language_context.py
```diff
@@ -3,6 +3,7 @@
 from aac.lang.active_context_lifecycle_manager import get_initialized_language_context
 from aac.lang.language_context import LanguageContext
 from aac.spec import get_aac_spec, get_primitives, get_root_keys
+from aac.plugins.validators.required_fields import get_required_fields
```
Coffee2Bits, line 6: Related to my other comment about moving the location of `get_required_fields`; The tests exercising the method `get_required_fields` should be in the corresponding test file. If you move the method out of the validator plugin, you can ignore this comment.

Coffee2Bits (changes requested):

python/src/aac/lang/definitions/extensions.py
```diff
@@ -106,18 +106,27 @@ def _remove_schema_fields(target_fields: list, fields_to_remove: list):
...
-    if extension_required_fields:
-        if "required" not in target_definition_fields:
-            target_definition_fields["required"] = []
+    if not definition_validations:
+        target_definition_fields["validation"] = []
 
-        target_definition_fields["required"] += extension_required_fields
+    required_fields_validation = _get_required_fields_validation_for_definition(target_definition_fields)
+    validation_names = _get_validation_names_for_definition(target_definition_fields)
+    if extension_required_fields and required_fields_validation not in validation_names:
```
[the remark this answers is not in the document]
cvchaparro: Good point; I'm not even sure I remember why I included that, anymore.

python/src/aac/plugins/gen_design_doc/gen_design_doc_impl.py
cvchaparro: For documentation purposes, here's my reasoning for why I moved that function out of the `Definition` class to begin with:

- I'd argue the logical place to put a `get_required_fields` function is in the place where required fields are added to the language, which means (in this case) in the validator plugin that handles required fields.

  Of course, that would mean I'd need to think more about the design because the validator plugin isn't currently responsible for adding/removing required fields so by this logic everything about required fields should be moved to the required fields plugin.

  *QUESTION*: Do we want a future story for doing this?
- If we leave it in the `Definition` class, we introduce a dependency into the core package on a plugin.

  This effectively means that changes to the validator plugin could cause changes to the core `aac` package which doesn't make sense, in my opinion. That kind of thing should go the other way.

  I think we should strive to have the core package follow the open/closed principle where the open part comes from plugins but if a change to a plugin causes the core package to change, that would be an undesirable change. Perhaps this indicates that we should update the design of validator plugins?

cvchaparro: Per our conversation, I'll resolve this and leave the function in-place for now.

python/tests/lang/test_language_context.py
Coffee2Bits: We talked about this. Going to resolve this comment.

Coffee2Bits (approved):
Looks good!

PR #358, opened 2022-06-06
Per python version test matrix

Coffee2Bits:
I had an early adopter reach out and ask about compatibility with Python version 3.10.4, so I went ahead and added a matrix for the python tests to include running in different python version environments. 

* Added a matrix for functional python tests
  * Python 3.9.0
  * Python 3.9.13
  * Python 3.10.4


.github/workflows/pull-requests.yml
```diff
@@ -6,15 +6,15 @@ on:
...
 jobs:
   build_python_artifacts:
-    name: Build Python Artifacts
-    uses: jondavid-black/AaC/.github/workflows/python-build-artifacts.yml@923464a8542c6cfa46018b45f50f1ffc455d5de9
+    name: Python Build
+    uses: jondavid-black/AaC/.github/workflows/python-build-artifacts.yml@5f36be5cb244cad6aaa226fb6cbae07a63ad9074
 
   build_vscode_artifacts:
-    name: Build VSCode Extension Artifacts
+    name: VSCode Ext. Build
```
cvchaparro, line 13:
```suggestion
    name: VSCode Extension Build
```

.github/workflows/python-tests.yml
```diff
@@ -8,34 +6,40 @@ defaults:
...
-    name: Python Package Functional Tests
+  python_tests:
+    name: Python ${{ matrix.py_version }} Tests
     runs-on: ubuntu-latest
 
-    outputs:
-      wheel_name: ${{ steps.wheel_artifact.outputs.wheel_name }}
+    strategy:
+      matrix:
+        py_version: ["3.9.0", "3.9.13", "3.10.4"]
```
cvchaparro, line 15: Per our previous discussion, I think we want this to be 3.10.5.

```suggestion
        py_version: ["3.9.0", "3.9.13", "3.10.5"]
```

.github/workflows/python-tests.yml
```diff
@@ -83,7 +104,7 @@ jobs:
             path: python/dist/
```
cvchaparro, line 104: (I can't leave a comment on lines 94 and 97, or I'd say this there) Also, should we update the 3.9 to use the python version matrix?

Coffee2Bits: I decided that hardcoding the value was the lesser evil when compared to duplicating the awk line, which will match the 3.9 value. If we expected to update that value, then I'd be inclined to duplicate the awk command.

.github/workflows/pull-requests.yml
Coffee2Bits: Spelling out Extension makes the step name unreadable in Github's interface. ☹️

PR #372, opened 2022-06-15
LSP Should handle malformed definitions when updating definitions

### LINKED ISSUE #331: I want the LSP to handle ingesting malformed and updated definitions
As an AaC user that writes AaC files, I want the LSP to handle ingesting malformed and updated definitions, so that the LSP can process definitions after I have altered them in my IDE.

relates to: #320 and #324 

Notes:
* Cameron raised the point that there may be value for us to use alternative events in the save lifecycle like `textDocument/willSave` and `textDocument/willSaveWaitUntil`
https://github.com/jondavid-black/AaC/pull/325#discussion_r866927874

* Definitions can be pruned by leveraging the definition's schema and removing any fields/sections that aren't defined in the schema, or may be misspelled.

## AC:
- [x] The LSP updates definitions on the `textDocument/didSave` notification
- [x] Definitions with fields that are not specified by their schemas have those fields removed before being ingested into the language context

Coffee2Bits:
Closes #331

* Added a function to strip undefined, or incorrectly labeled, fields from definitions
* LSP automatically strips incoming definitions for onChange events


python/src/aac/lang/lsp/language_server.py
```diff
@@ -93,29 +94,30 @@ async def did_close(ls: AacLanguageServer, params: DidCloseTextDocumentParams):
...
 
 
-async def did_change(ls: AacLanguageServer, params: DidChangeTextDocumentParams):
+async def did_save(ls: AacLanguageServer, params: DidSaveTextDocumentParams):
     """Text document did change notification."""
     logging.info(f"Text document altered by LSP client {params.text_document.uri}.")
 
-    file_content = params.content_changes[0].text
+    file_content = params.text
     incoming_definitions = parse(file_content)
```
cvchaparro, line 102: When I test locally and in Gitpod, I get the following error:

```
TypeError: lstat: path should be string, bytes or os.PathLike, not NoneType
```

which turns out to be a result of `params.text` being `None` when the `textDocument/didSave` request is made.

To demonstrate this issue, I opened the `aac/python/model/counter/counter.yaml` file and make a small change to spacing and saved. Again this is happening locally for me and in Gitpod. I think we could use `params.text_document.uri` slightly modified to not include the `file://` part (i.e. something like `params.text_document.uri.removeprefix("file://")`) to resolve what I'm seeing but it seems like this may have already been working for you so I'm open to discussing alternate solutions.

cvchaparro (changes requested):

python/src/aac/lang/lsp/language_server.py
Coffee2Bits: Implemented your suggestion. Good catch; Thanks!

cvchaparro (approved):
Nice work!

PR #376, opened 2022-06-24
Update the parser

### LINKED ISSUE #363: The parser should only populate lexemes and content relevant to the definition.
The parser should only populate lexemes and content relevant to the definition. As it stands, when multiple definitions are present in a YAML file, then each definition parsed from the YAML file will have its content set to the entire file and its lexemes will encompass the entire file.

## AC:
- [x] When a definition is parsed, its `lexemes` attribute only includes:
  - [x] A preceding logical document separator `---` 
  - [x] The text that makes up the definition's structure
- [x] When a definition is parsed, its `content` attribute only includes:
  - [x] A preceding logical document separator `---` 
  - [x] The text that makes up the definition's structure

cvchaparro:
Closes #363

Changes:
- Refactor the parser to build `Definition`s with only the `content` and `lexemes` that are related to the AaC definition that was parsed.


python/src/aac/parser/_parse_source.py
```diff
@@ -27,92 +31,62 @@ def parse(source: str, source_uri: Optional[str] = None) -> list[Definition]:
...
+        for root in _parse_yaml(source, doc):
+            if "import" in root:
+                del root["import"]
 
-        root_type, *_ = root.keys()
-        root_name = root.get(root_type).get("name")
-        parsed_models = parsed_models | {root_name: {source: root}}
+            root_type, *_ = root.keys()
+            root_name = root.get(root_type).get("name")
+            contents = _add_yaml_document_separator(doc)
```
Coffee2Bits, line 85: We can't assume that every definition has a preceding yaml doc separator. We should use regex to split the definitions without destructively removing yaml separators, however I wasn't able to quickly identify a regex that was really capable of accomplishing that.

Coffee2Bits (changes requested):
Looks good! Just a little feedback on the yaml-doc separators.

python/src/aac/parser/_parse_source.py
cvchaparro: I don't really see this assumption here so I'm not really sure I see the problem you're pointing out. If there is no document separator in the string/file contents we would add it (which is what was desired based on the AC). If there is no document separator and there should be one (i.e. there are two AaC definitions in the same YAML document) the user would eventually get an error to that effect since we're checking for that all over the place.

Can you give an example of a YAML document that would show what the issue is? In that way, at the very least I can create a test that confirms this works how we want it to.

Coffee2Bits: With the assumption you're errantly adding document separators to the initial definition of files like gen_protobuf.yaml. https://github.com/jondavid-black/AaC/blob/c8bc393ab7432d376f0c660a5de7a280329df10f/python/src/aac/plugins/gen_protobuf/gen_protobuf.yaml#L1-L43

As for the AC, the intention is to only add preceding doc separators that are present in the file, not to introduce content that's not present in the source file.

Coffee2Bits (approved):
Looks good, great job!

PR #377, opened 2022-06-24
Add template goto definition support and tests

### LINKED ISSUE #349: I want go to definition for the AaC DSL
As a user of the AaC LSP server, I want go to definition implemented, so that I can easily find the definition of items in the AaC language.

## AC:
- [x] Users can look up the definition of AaC definitions via the definition's name
- [x] Users can look up the root-key definitions via selecting the root key
- [x] When a user looks up an AaC definition via the LSP, the IDE opens an editor with the file containing the target `schema` definition

cvchaparro:
Closes #349

Changes:
- Implement "Goto Definition" for AaC definitions (i.e. `schema` definitions, `enum` definitions, and root definitions) in the LSP server.

Caveat:
This works for definitions regardless of whether they're in the same file being viewed by the user, or not, but if the definition is in a different file it currently must be opened in order for the "Goto Definition" action to work (this includes definitions defined in the core spec - i.e. `spec.yaml` must be opened to navigate to the `model` definition, etc). My understanding is that @Coffee2Bits is working on a file walker in a separate PR that I think will resolve this issue, so I'm not worrying about it here.


Coffee2Bits:
I was able to confirm that the goto definition works across files when the files are automatically opened via the VSCode extension keying off of the .aac file extension.

python/tests/lang/lsp/base_lsp_test_case.py
```diff
@@ -118,51 +117,127 @@ def to_uri(self, file_name: str) -> Optional[str]:
...
+        self.assertIsNotNone(
+            self.documents.get(file_name),
+            f"Could not execute {method} in virtual document because there is no document named {file_name}."
         )
-        return CompletionResponse(response.result())
+        response = await self.client.send_request(method, params)
+        return response_type(response.result())
+
+
+TEST_DOCUMENT_NAME = "test.aac"
```
Coffee2Bits, line 142: If we're going to consolidate constants for testing, let's move it into a constants file in the `tests/lang/lsp` or `test.helpers` package.

python/src/aac/lang/lsp/providers/goto_definition_provider.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def get_definition_location(self, documents: dict[str, Document], current_uri: str, position: Position) -> list[Location]:
+        """Return the location where the AaC definition is defined."""
+        name = documents.get(current_uri).word_at_position(position)
+        if not name:
+            return []
+
+        locations = []
+        for doc in documents.values():
+            ranges = self.get_ranges_containing_name(doc.source, name)
```
Coffee2Bits, line 26: I'm catching some errant definition goto suggestions, we should make this more explicit to avoid bad suggestions. For example, when I goto definition for `enum`, I get the schema definition for `enum`, the root key entry for `enum`, and the schema definition for `enumExt`. 

The cleaner behavior should be:
A. goto definition on a reference to a type goes to the type's schema definition.
B. goto definition on a root key goes to the root-key declaration.

Coffee2Bits (changes requested):
Looks good, but there is some minor feedback.

python/src/aac/lang/lsp/providers/goto_definition_provider.py
cvchaparro: Okay, I fixed this but I realized that I'm not handling enumerated value definitions correctly. I'll work on that tomorrow.

Coffee2Bits (changes requested):
Have a few more items, but it's looking really good!

python/src/aac/lang/lsp/providers/goto_definition_provider.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+
+        def get_end_position(position: Position) -> Position:
+            end_position = position.copy()
+            end_position.character += len(name)
+            return end_position
+
+        lines = content.splitlines()
+        starting_positions = [Position(line=i, character=lines[i].find(name)) for i, line in enumerate(lines) if name in line]
+        return [Range(start=start_pos, end=get_end_position(start_pos)) for start_pos in starting_positions]
```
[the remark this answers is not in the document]
cvchaparro: Per our discussion, I am going to address this in #390.

Coffee2Bits (approved):
Looks good!

PR #378, opened 2022-06-26
Rest API File Management

### LINKED ISSUE #371: I want the REST API Plugin to allow users to manage AaC files
As a user of the REST API, I want the REST API Plugin to allow users to manage AaC files, so that I can add, delete, modify, and perform other general management functions on AaC files in the context.

## AC:
- [x] Expose CRUD operations for AaC files in the REST API
   - [x] Create new AaC files
   - [x] Read existing AaC files
   - [x] Rename AaC files (File content updating is managed by Definition CRUD operations)
   - [x] Delete AaC files
- [x] The REST API uses the process' CWD as the workspace root URI
  - [x] Any AaC files in the root workspace or a sub-directory is automatically made available for importing to the context
  - [x] Sub-directories are captured and exposed via the REST API
- [x] The REST API exposes commands to add/remove AaC files from the context

Coffee2Bits:
Closes #371 

* Added a File class to represent AaC Files
* Moved Parser, Files, and File-writing module into a consolidated "io" package
* Added a filesystem walker to detect AaC files in a directory and its subdirectories
* Added file-management endpoints for the AaC REST API


python/src/aac/plugins/rest_api/aac_rest_app.py
```diff
@@ -1,99 +1,223 @@
...
+    return [to_file_model(file) for file in get_active_context().get_files_in_context()]
+
+
+@app.get("/files/available", status_code=HTTPStatus.OK, response_model=list[FileModel])
+def get_available_files(background_tasks: BackgroundTasks):
+    """Return a list of all files available in the workspace for import into the active context. The list of files returned does not include files already in the context."""
+    # Update the files via an async function so that any changes to the files shows up, eventually.
+    background_tasks.add_task(refresh_available_files_in_workspace)
+
+    #  Having to use a cached response for now as the file-walking makes the response take about 5 seconds, which is too long.
```
Coffee2Bits, line 35:
```suggestion
    #  Having to use a cached response for now as the file-walking makes the response take too long.
```

python/src/aac/io/writer.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    Given a list of definitions, write them to file uri. Updates the source for definitions passed in.
+
+    Args:
+        definitions (list[Definition]): The definitions to write to file.
+        file_uri (str): The URI of the file to write the definitions to.
+        is_user_editable (bool): True if the AaC file can be edited by users.
+    """
+    def sort_definitions_by_lexeme_line(definition: Definition) -> int:
+        return -1 if not definition.lexemes else definition.lexemes[0].location.line
```
cvchaparro, line 48: When would a definition not have lexemes? Is this mainly to account for definitions created for tests?

python/src/aac/lang/language_context.py
```diff
@@ -313,6 +314,38 @@ def get_definitions_by_file_uri(self, file_uri: str) -> list[Definition]:
...
+
+        return list(files_in_context.values())
+
+    def get_file_in_context_by_uri(self, uri: str) -> Optional[AaCFile]:
+        """
+        Return the AaCFile object by uri from the context or None if the file isn't in the context.
+
+        Args:
+            uri (str): The string uri to search for
+            language_context (LanguageContext): The language context to search for the file.
```
cvchaparro, line 342:
```suggestion
```

python/src/aac/plugins/rest_api/models/definition_model.py
```diff
@@ -1,27 +1,29 @@
...
 
+from aac.io.files.aac_file import AaCFile
 from aac.lang.definitions.definition import Definition
 
 
 class DefinitionModel(BaseModel):
     """REST API Model for the Definition class."""
     name: str
-    content: str
+    content: Optional[str] = ""
```
cvchaparro, line 12: If we're setting the default value to an empty string, do we really want this to be `Optional` (which would mean the default is `None`? I think we should pick one or the other, but using both seems incorrect.

python/src/aac/plugins/rest_api/models/file_model.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import Optional
+from pydantic import BaseModel, FilePath
+
+from aac.io.files.aac_file import AaCFile
+
+
+class FileModel(BaseModel):
+    """REST API Model for the File class."""
+    uri: str
+    content: Optional[str] = ""
```
cvchaparro, line 11: I have the same comment here, as in the `definition_model` module regarding `Optional`.

python/src/aac/plugins/rest_api/models/file_model.py
```diff
@@ -0,0 +1,55 @@
...
+
+    Returns:
+        A FileModel for use in the REST API.
+    """
+    return FileModel(
+        uri=file.uri, content=file_content, is_user_editable=file.is_user_editable, is_loaded_in_context=file.is_loaded_in_context
+    )
+
+
+def to_file_class(file: FileModel) -> AaCFile:
```
cvchaparro, line 43: Can we call this something like `to_aac_file`, instead?

python/tests/io_tests/test_paths.py
```diff
@@ -0,0 +1,41 @@
+import os
+from nose2.tools import params
+from unittest import TestCase
+
+from aac.io.paths import sanitize_filesystem_path
+
+WORKING_TEST_DIR = os.getcwd()
+
+
+def _get_working_directory_parent_directory(level_of_directories_up: int = 1) -> str:
```
cvchaparro, line 10: Is there a reason not to make this a general utility?

python/tox.ini
```diff
@@ -15,7 +15,7 @@ docstring-convention = google
 count = True
 per-file-ignores =
     # Ignore docstring errors in tests
-    *tests/*:D100,D101,D102,D104,D200,ANN101,ANN201,W293,W291,ANN001,E731
+    *tests/*:D100,D101,D102,D104,D200,ANN101,ANN201,W293,W291,ANN001,E731,W605
```
cvchaparro, line 18: Why are we ignoring this warning?

python/tests/io_tests/files/test_find.py
```diff
@@ -0,0 +1,117 @@
...
+                    l3_invalid_file_aac.close()
+
+        actual_result_file_paths = [file.uri for file in actual_result]
+        self.assertListEqual(expected_result, actual_result_file_paths)
+        self.assertNotIn(l1_invalid_file_aac.name, actual_result_file_paths)
+        self.assertNotIn(l2_invalid_file_yaml.name, actual_result_file_paths)
+        self.assertNotIn(l3_invalid_file_aac.name, actual_result_file_paths)
+
+
+def _write_content_to_temp_file(temp_file, content: str) -> None:
```
cvchaparro, line 88: Why use this instead of `temporary_test_file`? It seems like the places this is used should be able to be done using that helper.

python/src/aac/plugins/rest_api/aac_rest_app.py
```diff
@@ -1,22 +1,113 @@
...
+def get_files_from_context():
     """Return a list of all files contributing definitions to the active context."""
-    file_names = [definition.source_uri for definition in get_active_context().definitions]
-    return {"files": set(file_names)}
+    return [to_file_model(file) for file in get_active_context().get_files_in_context()]
+
+
+@app.get("/files/available", status_code=HTTPStatus.OK, response_model=list[FileModel])
+def get_available_files(background_tasks: BackgroundTasks):
+    """Return a list of all files available in the workspace for import into the active context. The list of files returned does not include files already in the context."""
```
cvchaparro, line 32: Something seems off, here, and I'm unsure if it's the docstring or the execution. As written, I'd expect this to return several files that can be imported into the context but the first time I make this request, I get an empty list as output (after starting the rest server from the `aac/python` directory).

This seems to be a result of the fact that the `refresh_available_files_in_workspace` function doesn't complete before returning because calling it again, I get the expected output. I'm not too familiar with `BackgroundTasks` but if it's an asynchronous function, do we need to `await` it so it completes before running?

Alternatively, feel free to point out what is wrong with my call.

python/src/aac/plugins/rest_api/aac_rest_app.py
```diff
@@ -1,22 +1,113 @@
...
+
+
+@app.post("/files/import", status_code=HTTPStatus.NO_CONTENT)
+def import_files_to_context(file_models: list[FilePathModel]):
+    """Import the list of files into the context."""
+    files_to_import = set([str(model.uri) for model in file_models])
+    valid_aac_files = set(filter(is_aac_file, files_to_import))
+    invalid_files = files_to_import.difference(valid_aac_files)
+
+    if len(invalid_files) > 0:
```
cvchaparro, line 62: I think we should import valid files, even if there are invalid ones.

cvchaparro (changes requested):
I have a few comments/suggestions.

python/src/aac/io/writer.py
Coffee2Bits: While the normal flow usually ensures that lexemes are present, there is no step or function that ensures it so we can't assumes that lexemes will be present.

python/src/aac/io/paths.py
```diff
@@ -0,0 +1,57 @@
+"""Module for shared path functions."""
+
+import logging
+import os
+import unicodedata
+from urllib.parse import unquote
+
+
+def sanitize_filesystem_path(file_path: str) -> str:
+    r"""
```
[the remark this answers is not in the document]
Coffee2Bits: It's a flake rule. Removing the regular expression leads flake8 to report the following findings:
```
./src/aac/io/paths.py:10:1: D301 Use r""" if any backslashes in a docstring
    """
    Sanitizes user-provided file path strings to prevent injection and tampering of filesystem paths.

    This function does not determine if a user can write to the path location, it only sanitizes and evaluates the
    path to a full path. The user must still check that the path is a valid destination.

    This function performs the following sanitizing actions:

        * Removes any URL encoding
        * Converts the path to an ASCII encoding
        * Replaces instances of / or \ with os-specific path separators
        * Normalizes the path
        * Determines the realpath if symbolic links are used
        * Ensures that the path is an absolute path

    Path Traversal Security References:

        * OWASP https://owasp.org/www-community/attacks/Path_Traversal

    Args:
        file_path (str): The file path to sanitize

    Returns:
        A sanitized filepath.
    """
^
./src/aac/io/paths.py:20:38: W605 invalid escape sequence '\ '
    """
    Sanitizes user-provided file path strings to prevent injection and tampering of filesystem paths.

    This function does not determine if a user can write to the path location, it only sanitizes and evaluates the
    path to a full path. The user must still check that the path is a valid destination.

    This function performs the following sanitizing actions:

        * Removes any URL encoding
        * Converts the path to an ASCII encoding
        * Replaces instances of / or \ with os-specific path separators
        * Normalizes the path
        * Determines the realpath if symbolic links are used
        * Ensures that the path is an absolute path

    Path Traversal Security References:

        * OWASP https://owasp.org/www-community/attacks/Path_Traversal

    Args:
        file_path (str): The file path to sanitize

    Returns:
        A sanitized filepath.
    """
       
                             ^
1     D301 Use r""" if any backslashes in a docstring
1     W605 invalid escape sequence '\ '
2
```

python/src/aac/plugins/rest_api/models/definition_model.py
Coffee2Bits: Good catch! Removed the default value.

python/src/aac/io/files/find.py
```diff
@@ -0,0 +1,42 @@
...
+
+    return aac_files
+
+
+def is_aac_file(filepath: str) -> bool:
+    """Test if a target file is considered a valid AaC file."""
+    filename, file_ext = os.path.splitext(filepath)
+
+    is_valid_aac_file = False
+    if file_ext == ".yaml" or file_ext == ".aac":
```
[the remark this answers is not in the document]
Coffee2Bits: They're defined as constants in the test files. I was on the edge about rearranging all of that since they were already a number of other changes in this PR

python/src/aac/io/files/find.py
```diff
@@ -0,0 +1,42 @@
...
+        try:
+            parse(filepath)
+        except ParserError as error:
+            logging.info(f"File '{filepath}' is not a valid AaC file. Reason: {error.errors}")
```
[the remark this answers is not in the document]
Coffee2Bits: ParserError.errors is a list of strings. We can write a story to standardize/modify these errors and their structures.

python/src/aac/plugins/rest_api/aac_rest_app.py
Coffee2Bits: I added a line `asyncio.run(refresh_available_files_in_workspace())` to the server start up function, so the initial value won't be an empty list, however, it does take noticeably longer to start up. I think that issue is because the `is_aac_file` check is parsing the file. We could speed up that method by providing a quick parse that doesn't calculate and set lexemes.

python/src/aac/plugins/rest_api/aac_rest_app.py
Coffee2Bits: Committing partial state alterations based from an invalid request is not a good pattern; it'll complicate error handling and batching requests for clients.

python/src/aac/plugins/rest_api/models/file_model.py
Coffee2Bits: I would like to keep the to/from conversion functions consistent between the models, and `to_definition` is confusing, so I elected to delineate the internal class from the models by referring to the former as class and later as model.

python/tests/io_tests/test_paths.py
Coffee2Bits: I elected not to since it felt like a very special use case/need.

python/src/aac/plugins/rest_api/models/file_model.py
Coffee2Bits: Good catch

python/src/aac/io/files/find.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def find_aac_files(root_directory_to_search: str) -> list[AaCFile]:
+    """Return a list of paths to AaC files in or under the root directory."""
+    aac_files = []
+
+    root_dir_abs_path = os.path.abspath(root_directory_to_search)
+    if os.path.isdir(root_dir_abs_path):
+        directory_results = os.walk(root_dir_abs_path)
+        for directory_path, directory_names, filenames in directory_results:
```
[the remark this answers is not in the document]
Coffee2Bits: Applied suggestion

python/src/aac/io/files/find.py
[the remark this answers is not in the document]
cvchaparro: I'd support a new story for standardizing that; however, my suggestion was more geared towards the output of this message since it will show a list of strings which can make the output harder to read. I think the solution would be to join all the error strings with newlines into a single string and output that.

python/src/aac/io/paths.py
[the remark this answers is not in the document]
cvchaparro: Oh, interesting. It's not a huge deal now that I know but it might be better to say "normalize path separators" or something so we can avoid that.

python/tests/io_tests/files/test_find.py
cvchaparro: I think the way to get multiple CMs in the same statement is to use parentheses - at least that's how I've made it work previously.

python/src/aac/plugins/rest_api/aac_rest_app.py
cvchaparro: Fair enough. I was thinking that since (I'm pretty sure) adding definitions, etc is idempotent it wouldn't be a big deal but I guess there's a good argument for atomic operations, too.

cvchaparro (changes requested):
I left a few comments and resolved others. Also, we will need to resolve merge conflicts.

python/src/aac/io/files/find.py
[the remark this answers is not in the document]
Coffee2Bits: I think the log handles lists pretty well already. This is what the output looks like in the log:
```
07/18/2022 20:39:00::find::find.py::L37::INFO::File '/workspace/AaC/python/src/aac/main-branch.yaml' is not a valid AaC file. Reason: ["Failed to parse file: 'str' object has no attribute 'get'", '# This workflow is only intended to run when a pull-request is merged into the main branch\nname: Main branch Workflow\n\non:\n  push:\n    branches: [main]\n\njobs:\n  build_python_artifacts:\n    name: Python Build\n    uses: ./.github/workflows/python-build-artifacts.yml\n\n  build_vscode_artifacts:\n    name: VSCode Extension Build\n    needs: build_python_artifacts\n    uses: ./.github/workflows/vscode-build-artifacts.yml\n\n  python_tests:\n    name: Python Tests\n    needs: build_python_artifacts\n    uses: ./.github/workflows/python-tests.yml\n']
```

python/tests/io_tests/files/test_find.py
Coffee2Bits: I updated the test with the parentheses. Not sure why it didn't work the first time I tried with parentheses. 🤷‍♂️

python/src/aac/io/parser/_parse_source.py
```diff
@@ -31,7 +30,12 @@ def parse(source: str, source_uri: Optional[str] = None) -> list[Definition]:
         A list of Definition objects containing the internal representation of the definition and metadata
         associated with the definition.
     """
-    return _parse_file(source) if path.lexists(source) else _parse_str(source_uri or DEFAULT_SOURCE_URI, source)
+    sanitized_source = sanitize_filesystem_path(source)
+    return (
+        _parse_file(sanitized_source)
+        if path.lexists(sanitized_source)
+        else _parse_str(source_uri or DEFAULT_SOURCE_URI, source)
```
cvchaparro, line 37:
```suggestion
        else _parse_str(sanitize_filesystem_path(source_uri) or DEFAULT_SOURCE_URI, source)
```

cvchaparro (changes requested):

python/src/aac/io/files/find.py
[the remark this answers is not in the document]
Coffee2Bits: I'd agree if this wasn't for logging. Multi-line comments do make programmatic parsing of logs ever so slightly more difficult.

python/src/aac/io/parser/_parse_source.py
```diff
@@ -31,7 +30,12 @@ def parse(source: str, source_uri: Optional[str] = None) -> list[Definition]:
         A list of Definition objects containing the internal representation of the definition and metadata
         associated with the definition.
     """
-    return _parse_file(source) if path.lexists(source) else _parse_str(source_uri or DEFAULT_SOURCE_URI, source)
+    sanitized_source = sanitize_filesystem_path(source)
+    return (
+        _parse_file(sanitized_source)
+        if path.lexists(sanitized_source)
+        else _parse_str(sanitize_filesystem_path(source_uri) or DEFAULT_SOURCE_URI, source)
```
cvchaparro, line 37:
```suggestion
        else _parse_str(sanitize_filesystem_path(source_uri or DEFAULT_SOURCE_URI), source)
```

python/src/aac/io/files/find.py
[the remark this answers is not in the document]
cvchaparro: Per our discussion, we will aim to make logs easily machine-parseable so we can (eventually) create a log viewer that will alleviate some of the readability issues, I'm fine leaving this as-is.

cvchaparro (approved):

Issue #375, opened 2022-06-23
Model references


jondavid-black:
As a modeler, I want to be able to reference information across different models so that I can create more complex relational data representations in my models.

AC:

- [x] Must create a primitive type for references.
- [x] Must create a reference format validator.
- [x] Must create a reference existence validator. 

Initial thoughts on reference format:

- Use "dot notation" from a root type to indicate navigation of a field hierarchy
- In many instances a pure "dot notation" will be ambiguous, so provide a means to select specific data instances.  The simple way to do this is to match a field name with a specified value.  The notation may look like `[name=myModel]`.
- The reference format validator should just be a regex ensuring proper "dot notation" and selector syntax was used.  There may be instances where a reference is needed but the target item is not defined.
- The reference existence validator should search the model for an instance of the reference target and ensure it exists.

The initial reference should only be a single field definable in a schema.  In the future we may wish to explore use of inline references to support things like input references within a behavior test specification.  That concept is intentionally omitted from the AC at this time to limit complexity.

jondavid-black:
This appears to be working but needs some refactoring and testing before I'm comfortable with it.

PR #379, opened 2022-06-28
375 model references

jondavid-black:
Pull request for #375 to add model reference primitive and associated validation.


python/src/aac/lang/references.py
```diff
@@ -21,3 +22,198 @@ def filter_definitions_by_reference(definition_to_filter: Definition) -> bool:
...
+                if isinstance(item, dict):
+                    i = _drill_into_nested_dict(search_keys[1:], item)
+                    if len(i) > 0:
+                        items_to_return.extend(i)
+            return items_to_return
+        elif isinstance(next_level, dict):
+            return _drill_into_nested_dict(search_keys[1:], next_level)
+        else:
+            return []
+    return []
```
Coffee2Bits, line 202: I was able to simplify this a touch by leveraging the dict function `get` which returns the value for the key or None, and that the result of the recursive function call will be a list.

```suggestion
    items_to_return = []
    if search_keys[0] in search_me.keys():
        next_level = search_me.get(search_keys[0])
        if isinstance(next_level, list):
            # iterate through next level items and continue search
            for item in next_level:
                items_to_return.extend(_drill_into_nested_dict(search_keys[1:], item))
        elif isinstance(next_level, dict):
            items_to_return = _drill_into_nested_dict(search_keys[1:], next_level)

    return items_to_return
```

python/src/aac/lang/references.py
```diff
@@ -21,3 +22,198 @@ def filter_definitions_by_reference(definition_to_filter: Definition) -> bool:
...
+                    else:
+                        if str(value_for_key[selector_field]) == str(selector_value):  # casting these seems like a hack
+                            keepers.append(definition)
+                else:
+                    # if there's no selector, keep the definition since it contains the segment key
+                    keepers.append(definition)
+    return keepers
+
+
+def _drill_into_nested_dict(search_keys: list[str], search_me: dict) -> list:
```
Coffee2Bits, line 183:
```suggestion
def _drill_into_nested_dict(search_keys: list[str], search_me: dict) -> list[dict]:
```

python/src/aac/lang/references.py
```diff
@@ -21,3 +22,198 @@ def filter_definitions_by_reference(definition_to_filter: Definition) -> bool:
...
+                    value_for_key = def_dict[key]
+                    if type(value_for_key) is list:
+                        for item in value_for_key:
+                            if str(item[selector_field]) == str(selector_value):  # casting these seems like a hack
+
+                                keepers.append(definition)
+                                break
+                    else:
+                        if str(value_for_key[selector_field]) == str(selector_value):  # casting these seems like a hack
+                            keepers.append(definition)
```
Coffee2Bits, line 176: It looks like you should be able to remove the string casting -- the tests still pass after I removed them. 

I'm happy to help identify/test a scenario in which we're expecting non-string values in the comparisons.
```suggestion
                    if isinstance(value_for_key, list):
                        for item in value_for_key:
                            if item.get(selector_field) == selector_value:
                                keepers.append(definition)
                                break
                    elif value_for_key.get(selector_field) == selector_value:
                        keepers.append(definition)
```

python/src/aac/lang/references.py
```diff
@@ -21,3 +22,198 @@ def filter_definitions_by_reference(definition_to_filter: Definition) -> bool:
...
+
+
+def _has_disallowed_characters(test_me: str) -> bool:
+    disallowed = set("~`!@#$%^&*()=+\\|[]{}'\";:/?,.<> ")
+    return any((c in disallowed) for c in test_me)
+
+
+def get_reference_target_definitions(reference_field_value: str, language_context: LanguageContext) -> list[Definition]:
+    """
+    Return a boolean indicating if the reference target exists in the defined model.
```
Coffee2Bits, line 85:
```suggestion
    Return a list containing the referenced definitions.
```

python/src/aac/lang/references.py
```diff
@@ -21,3 +22,198 @@ def filter_definitions_by_reference(definition_to_filter: Definition) -> bool:
...
+    return _process_segment(nxt_prefix, segments[1:], keepers, language_context)
+
+
+def _process_root(selector: str, selector_field: str, selector_value: str, definitions: list[Definition]) -> list[Definition]:
+    """Return definitions that match the root selector.  If no selector, return all definitions."""
+    keepers = []
+    if selector:
+        # handle root level selector
+        for definition in definitions:
+            if str(definition.get_top_level_fields()[selector_field]) == str(selector_value):
```
Coffee2Bits, line 147: The tests indicate that you can get away without casting here. I'm happy to help identify/test some additional scenarios which would demonstrate the need to cast these values.
```suggestion
            if definition.get_top_level_fields().get(selector_field) == selector_value:
```

Coffee2Bits (changes requested):
Looks good! I like what you've got and where you're going with these references.

I've left some feedback as suggestions. I'm happy to discuss them if you have questions/concerns.

python/src/aac/lang/references.py
jondavid-black: I put that cast in because I ran into some hard to diagnose python errors while I was doing development.  It may have become a non-issue as I worked out some bugs, but I'm still hesitant to make to make the change.  As a compromise, I'll add a new issue to expand the test cases and remove casts if they all pass.  I think the issue may have occurred when the field was an integer.  Now I realize I didn't explicitly create a unit test for that condition.

python/src/aac/lang/references.py
jondavid-black: I put that cast in because I ran into some hard to diagnose python errors while I was doing development.  It may have become a non-issue as I worked out some bugs, but I'm still hesitant to make to make the change.  As a compromise, I'll add a new issue to expand the test cases and remove casts if they all pass.  I think the issue may have occurred when the field was an integer.  Now I realize I didn't explicitly create a unit test for that condition.

PR #388, opened 2022-07-07
I want the LSP Hover Request to show the structure of a definition when I hover over its name

cvchaparro:
As an AaC user of the LSP, I want the LSP Hover Request to show the structure of a definition when I hover over its name, so that I can access the structure of a definition on-demand.

## AC: 
- [x] When the user hovers over the name of a definition in the context, the definition's structure is returned as the hover response content


cvchaparro:
This should be merged after #377 as I branched from that.

Coffee2Bits:
I was having a lot of fun testing this out, but I noticed that the window is a little small, but this seems like a long-lived issue in VSCode: https://github.com/microsoft/vscode/issues/14165

Coffee2Bits (approved):
Looks good!

PR #396, opened 2022-07-22
Alexcmonnet/i-want-the-rest-api-351

### LINKED ISSUE #351: I want the REST API to expose all AaC commands

Coffee2Bits:
Closes #351

* Added REST endpoint for getting available plugin commands
* Added REST endpoint for executing AaC plugin commands
* Added filter to prevent long-running commands from being executed via the REST API


cvchaparro (changes requested):
Nice work, overall. One thing I did notice, though, is that commands that should fail, don't seem to. Specifically, I ran `validate` on the `alarm_clock.yaml` file after making changes such that it should be invalid (i.e. changing `Clock` to `Klock` and removing `when` from the scenario) and it still returned a valid response/message. This doesn't seem to be an issue when running from the command line, however.

python/src/aac/plugins/rest_api/models/command_model.py
```diff
@@ -0,0 +1,46 @@
...
+class CommandRequestModel(BaseModel):
+    """Command model for user requests to execute commands."""
+    name: str
+    arguments: list[str]
+
+
+class CommandResponseModel(BaseModel):
+    """Response model for returning the results/outcomes of commands requests."""
+    command_name: str
+    result_message: str
```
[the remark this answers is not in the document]
Coffee2Bits: Good call out!

python/src/aac/plugins/rest_api/models/command_model.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class CommandRequestModel(BaseModel):
+    """Command model for user requests to execute commands."""
+    name: str
+    arguments: list[str]
+
+
+class CommandResponseModel(BaseModel):
+    """Response model for returning the results/outcomes of commands requests."""
```
Coffee2Bits, line 14:
```suggestion
    """Response model for returning the results/outcomes of command requests."""
```

python/src/aac/plugins/rest_api/models/command_model.py
```diff
@@ -0,0 +1,47 @@
...
+    command_name: str
+    result_message: str
+    success: bool
+
+
+class CommandArgumentModel(BaseModel):
+    """REST API Model for the AaC Command class."""
```
Coffee2Bits, line 21:
```suggestion
    """REST API Model for the AaC Command Argument class."""
```

cvchaparro (changes requested):
After our conversation about restarting the Rest API to test unsuccessful commands, I was able to confirm everything was working but we should probably have some kind of statement somewhere saying that it's required since, without thinking about it and knowing that the language context won't update definitions that are already present, I could see that being confusing and leading to erroneous results.

The last thing I'd point out is that `aac_rest_app.py` and `test_rest_api.py` seem to have been reformatted (at least partially) due to, what I assume is the result of a mismatched line length. Can you fix that - i.e. revert those formatting changes - please?

Coffee2Bits:
> After our conversation about restarting the Rest API to test unsuccessful commands, I was able to confirm everything was working but we should probably have some kind of statement somewhere saying that it's required since, without thinking about it and knowing that the language context won't update definitions that are already present, I could see that being confusing and leading to erroneous results.
> 
> The last thing I'd point out is that `aac_rest_app.py` and `test_rest_api.py` seem to have been reformatted (at least partially) due to, what I assume is the result of a mismatched line length. Can you fix that - i.e. revert those formatting changes - please?

I've fixed the formatting. The files are currently formatted per the black settings. Confirmed no change when black was run against the files `python -m black --line-length=127 --diff --quiet ./tests/plugins/test_rest_api.py`

cvchaparro (approved):
Nice work!

PR #403, opened 2022-07-26
Alexcmonnet/validating yaml with 386

### LINKED ISSUE #386: Validating yaml with malformed ext does not work properly
I've been beating my head against the validator for a while now.  All it was providing was a general error and a line reference nowhere close to where the issue was.  It appears the language context is trying to apply the `ext` definition in the model file being validated rather than just validate the contents.

To repeat the issue define a enum incorrectly (in this case using the old `dataExt` rather than `schemaExt`):
```
ext:
  name: addReportToRoot
  type: root
  dataExt:
    add:
      - name: report
        type: Report
```
When you run `aac validate file-with-invalid-aac.yaml` you get the following resopnse:
```
validate: general_failure

An unrecognized error occurred:/workspace/AaC/python/src/aac/cli/builtin_commands/validate/validate_impl.py line 20 message: 'NoneType' object has no attribute 'get'
```
The expected behavior would be a message stating that a defined field in the provided yaml file is incorrect.

Coffee2Bits:
Closes #386

* Added error handling for malformed extensions

### DIFF
--- a/python/src/aac/validate/_validate.py
+++ b/python/src/aac/validate/_validate.py
@@ -3,6 +3,7 @@ from iteration_utilities import flatten
 from typing import Generator
 
 from aac.lang.language_context import LanguageContext
+from aac.lang.language_error import LanguageError
 from aac.lang.active_context_lifecycle_manager import get_active_context
 from aac.lang.definitions.definition import Definition
 from aac.lang.definitions.schema import get_definition_schema_components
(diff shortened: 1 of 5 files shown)


python/src/aac/validate/_validate.py
```diff
@@ -43,10 +44,13 @@ def validated_source(source: str) -> Generator[ValidationResult, None, None]:
 
 def _with_validation(user_definitions: list[Definition]) -> ValidationResult:
     result = _validate_definitions(user_definitions)
-    if result.is_valid:
-        return result
-    else:
-        raise ValidationError("Failed to validate content with errors:", *result.messages)
+    try:
```
cvchaparro, line 47: I couldn't leave it as a suggestion but I think this `try` should go around the `result = ...` part, too. I don't see any way in which this `if` statement would result in a `LanguageError`.

cvchaparro (changes requested):
Looks good, there was just one thing I noticed that I think needs to be changed.

python/src/aac/validate/_validate.py
Coffee2Bits: Good catch, thanks!

cvchaparro (approved):
Nice work!

Issue #313, opened 2022-04-24
I want the help-dump output sorted for consistent results


Coffee2Bits:
I want the help-dump output sorted alphabetically for consistent results.

Notes:
The output should be filtered in the `_get_all_commands` function
https://github.com/jondavid-black/AaC/blob/45255db59bea22970200f46a610a02b3372acb60/python/src/aac/plugins/help_dump/help_dump_impl.py#L33

## AC:
- [ ] The aac commands are sorted alphabetically by command name

PR #411, opened 2022-07-27
Added in help_dump sorting for the output JSON dump

JustAChucklin:
This PR is in relation to this issue -> [Issue #313](https://github.com/jondavid-black/AaC/issues/313)
Added a sort function to the help_dump results. 

The results should be sorted by alphabetical order now. 

Also added a test case for this, and as of the last testing that was done the test was passing. 

Let me know if there is anything else that is needing to be added/changed regarding this item

### DIFF
--- a/python/src/aac/plugins/help_dump/help_dump_impl.py
+++ b/python/src/aac/plugins/help_dump/help_dump_impl.py
@@ -30,4 +30,4 @@ def help_dump() -> PluginExecutionResult:
 def _get_all_commands() -> list[AacCommand]:
     all_plugins = plugin_manager.get_plugin_manager().get_plugins()
     list_plugins = list(flatten([plugin.get_commands() for plugin in all_plugins if hasattr(plugin, "get_commands")]))
-      return sorted(list_plugins)
+    return sorted(list_plugins)
--- a/python/src/aac/plugins/help_dump/help_dump_impl.py
+++ b/python/src/aac/plugins/help_dump/help_dump_impl.py
@@ -27,7 +27,7 @@ def help_dump() -> PluginExecutionResult:
     with plugin_result(plugin_name, get_command_info) as result:
         return result
 
-
 def _get_all_commands() -> list[AacCommand]:
     all_plugins = plugin_manager.get_plugin_manager().get_plugins()
-    return list(flatten([plugin.get_commands() for plugin in all_plugins if hasattr(plugin, "get_commands")]))
+    list_plugins = list(flatten([plugin.get_commands() for plugin in all_plugins if hasattr(plugin, "get_commands")]))
+    return sorted(list_plugins)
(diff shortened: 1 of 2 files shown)


python/tests/plugins/test_help_dump_impl.py
```diff
@@ -37,6 +37,10 @@ def test_help_dump_prints_first_party_commands(self):
                 result.get_messages_as_string()
             )
 
+    def test_help_dump_json_dump_sort(self):
+        list_all_plugins = help_dump().messages
+        json.loads("".join(list_all_plugins), object_pairs_hook=OrderedDict)
+
```
Coffee2Bits, line 43: Here is a suggested change that asserts that command names in the help dump output are sorted alphabetically. 
```suggestion
        expected_commands = sorted(get_plugin_commands(), key=lambda command: command.name)

        list_all_plugins = help_dump().messages
        help_dump_output = json.loads("".join(list_all_plugins), object_pairs_hook=OrderedDict)

        expected_result = [command.name for command in expected_commands]
        actual_result = [command.get("name") for command in help_dump_output]

        self.assertListEqual(expected_result, actual_result)
```

python/src/aac/plugins/help_dump/help_dump_impl.py
```diff
@@ -30,4 +30,5 @@ def get_command_info():
 
 def _get_all_commands() -> list[AacCommand]:
     all_plugins = plugin_manager.get_plugin_manager().get_plugins()
-    return list(flatten([plugin.get_commands() for plugin in all_plugins if hasattr(plugin, "get_commands")]))
+    list_plugins = list(flatten([plugin.get_commands() for plugin in all_plugins if hasattr(plugin, "get_commands")]))
+    return sorted(list_plugins)
```
Coffee2Bits, line 34: Glad that it sorts by the name attribute by default, but I'd like to make it explicit for future readers. 

```suggestion
    return sorted(list_plugins, key=lambda command: command.name)
```

Coffee2Bits (changes requested):
I have a couple suggested changes for readability. You should be able to batch add them from Github. Once the pipeline passes I'll be happy to approve the PR. Good work!

python/tests/plugins/test_help_dump_impl.py
JustAChucklin: `expected_commands = sorted(get_plugin_commands()...` is throwing an unresolution error. I am testing locally for a fix

JustAChucklin: Ok got it working locally...the test that is...now the test is failing so this may need to be looked into. There are differences between the expected and the actual results

python/src/aac/plugins/help_dump/help_dump_impl.py
```diff
@@ -30,4 +30,5 @@ def get_command_info():
 
 def _get_all_commands() -> list[AacCommand]:
     all_plugins = plugin_manager.get_plugin_manager().get_plugins()
-    return list(flatten([plugin.get_commands() for plugin in all_plugins if hasattr(plugin, "get_commands")]))
+    list_plugins = list(flatten([plugin.get_commands() for plugin in all_plugins if hasattr(plugin, "get_commands")]))
+    return sorted(list_plugins, key=lambda command: command.name)
```
cvchaparro, line 34:
```suggestion
    plugins_with_commands = list(flatten([plugin.get_commands() for plugin in all_plugins if hasattr(plugin, "get_commands")]))
    return sorted(plugins_with_commands, key=lambda command: command.name)
```

Coffee2Bits (approved):

PR #416, opened 2022-08-09
Move the LSP server to a plugin and fix plugin_manager

### LINKED ISSUE #407: Update the Plugins and Plugin interfaces to use contribution points
Update the Plugins and Plugin interfaces to use contribution points. The `plugin_manager` should exclusively use the `get_plugins` hook.

### AC:
- [x] The only hook remaining is `get_plugins`
- [x] `plugin_manager` is update to only use the `get_plugins` hook
- [x] `plugin_manager` provides a function to get all plugin commands
- [x] `plugin_manager` provides a function to get all plugin definitions
- [x] `plugin_manager` provides a function to get all plugin validations
- [x] Plugins are update to only provide the `get_plugins` hook
- [x] Gen-Plugin plugin templates are updated appropriately

### LINKED ISSUE #410: Move the LSP server to a plugin
Currently, the LSP isn't setup as a plugin, but rather it's been stored incorrectly as a core package. 

## AC:
- [x] The LSP package is moved to an LSP plugin
- [x] LSP tests are updated accordingly

cvchaparro:
Closes #410
Closes #407

Changes:
- Move the all the LSP functionality into a plugin so that AaC is decoupled from editor-specific code.
- Remove all hookspecs except for the `get_plugin` hook.
- Update `plugin_manager` such that it uses the `get_plugin` hook in all functions.
- Update all plugins so that they implement the `get_plugin` hook and do not implement any of the old hooks.
- Update `gen-plugin` to generate plugins using only the `get_plugin` hook.


python/src/aac/plugins/validators/_common.py
```diff
@@ -16,12 +19,30 @@ def get_validation_definition_from_plugin_definitions(plugin_definitions_string:
...
     if len(validation_definitions) != 1:
         raise PluginError(
             f"Expected one and only one validation definition.\nValidation Definitions:\n{validation_definitions}"
         )
 
     return validation_definitions[0]
+
+
+def get_plugin_validations_from_definitions(plugin_definitions: list[Definition], callback: Callable) -> list[ValidatorPlugin]:
+    """Return the ValidatorPlugins based on the plugin definitions."""
```
Coffee2Bits, line 46: I'm conflicted on this. The function name and return type both imply a list of validator plugins, but until we implement the ability to support multiple validators per plugin this function could easily lead users to incorrectly believe that they can implement multiple validators per plugin.

This note in the docstring should suffice for now, unless you'd prefer to alter the return type and use the following pattern in the plugins:
```python
plugin.register_validations([_get_plugin_validation()])
```


```suggestion
    """
    Return the ValidatorPlugins based on the plugin definitions.
    
    Beware, plugins currently only support one validator per plugin. 
    """
```

Coffee2Bits (changes requested):
Great work! Just one open-ended suggestion/change.

python/src/aac/plugins/validators/_common.py
```diff
@@ -16,12 +19,34 @@ def get_validation_definition_from_plugin_definitions(plugin_definitions_string:
...
 
     return validation_definitions[0]
+
+
+def get_plugin_validations_from_definitions(plugin_definitions: list[Definition], callback: Callable) -> list[ValidatorPlugin]:
+    """
+    Return the ValidatorPlugins based on the plugin definitions.
+
+    Beware, plugins currently only support one validator per plugin. 
+    """
```
cvchaparro, line 50:
```suggestion
    """
    Return the ValidatorPlugins based on the plugin definitions.

    Beware, plugins currently only support one validator per plugin.
    """
```

Coffee2Bits (approved):
Looks good!

PR #419, opened 2022-08-15
Extend the VSCode tests timeout

cvchaparro:
The AaC tool's startup time has grown enough to where tests fail somewhat
consistently due to timeouts to increase the timeout until we can decrease the
startup time.


Coffee2Bits (approved):
👍

PR #423, opened 2022-08-18
Gen-Protobuf - Include spaces in the filter for the filenames

### LINKED ISSUE #342: Gen Protobuf Filenames should have spaces stripped
Gen Protobuf Filenames should have spaces stripped.

Example model that'll expose the issue:
```yaml
schema:
  name: User Request
  description: Example user request for blog post(s)
  fields:
    - name: UserID
      type: int
    - name: Blog Post IDs
      type: int[]
      description: The list of requested blog posts
```

JustAChucklin:
This closes #342 

This includes changes to the filter for Gen-Protobuf that adds the ability to strip spaces from the filenames. Also including a unit test for validation. 

Let me know if there is anything additional required for this. 

Thanks in advance!

### DIFF
--- a/python/tests/plugins/test_gen_protobuf.py
+++ b/python/tests/plugins/test_gen_protobuf.py
@@ -35,7 +35,6 @@ SCHEMA_FILE_TYPE = "schema"
 ENUM_FILE_TYPE = "enum"
 ENUM_VALUE_1 = "VAL1"
 ENUM_VALUE_2 = "VAL2"
-TEST_MESSAGE_SPACE_NAME = "Message A"
 
 
 class TestGenerateProtobufPlugin(ActiveContextTestCase):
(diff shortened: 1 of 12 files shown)


python/tests/plugins/test_gen_protobuf.py
```diff
@@ -138,6 +139,15 @@ def test_gen_protobuf_fails_with_multiple_message_templates(self, load_templates
     def test__convert_camel_case_to_snake_case(self, test_string, expected_string):
         self.assertEqual(expected_string, _convert_camel_case_to_snake_case(test_string))
 
+    @params(
+        ("Data A", "data_a.proto"),
+        ("Message with multiple spaces", "messagewithmultiplespaces.proto"),
```
Coffee2Bits, line 144: I'm not sure how these two tests don't conflict. Bear with me as I dig in a bit.

JustAChucklin: I was thinking of combining the test....but I wasnt sure if that would be ok to do or not.

Coffee2Bits: No, let's keep them separate as they demonstrate different test cases.

The disconnect is because of the snake case function which delineates snake case words by capitalization. So, the input of `Message With Multiple Spaces` would yield `message_with_multiple_spaces.proto`.

Coffee2Bits (approved):
Looks good! 👍

PR #427, opened 2022-08-26
Opening up Reference Field Validation to be more Lenient

JustAChucklin:
Updated the field_references validation to be more lenient and to go based on the arguments that are passed in from the YAML instead of the hard-coded 'type'. Also added in a loop to loop through all the types to make sure they are all valid types/fields. 

Updated the test for the reference fields as well since they were failing after the tests. 
Let me know if there is anything else needed for this or if there is anything that is needing to be changed.

### DIFF
--- a/python/src/aac/plugins/validators/defined_references/_validate_references.py
+++ b/python/src/aac/plugins/validators/defined_references/_validate_references.py
@@ -19,22 +19,27 @@ def validate_references(definition_under_test: Definition, target_schema_definit
         A ValidatorResult containing any applicable error messages.
     """
     error_messages = []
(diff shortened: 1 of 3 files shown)


JustAChucklin:
Two things that I am currently implementing a solution for. I am currently trying to get the test to pass as the solution that was implemented first is failing the tests. 

I am also thinking that I can expand the current code that is there to include names and that should be sufficient to expand this validate

JustAChucklin:
Opening this up as with the suggestion from @Coffee2Bits  the tests are now passing.

Coffee2Bits (approved):
Looks good!

PR #439, opened 2022-09-12
Update the LanguageContext to use UUIDs as ultimate internal identifiers instead of names

cvchaparro:
Update the LanguageContext to use UUIDs as ultimate internal identifiers instead of names, so that users can rename definitions.

## AC:
- [x] Definition class is updated to include a private-ish field for a UUID
- [x] The Language context's dictionary "definitions_name_dictionary" is updated to use definition UUID's as keys.
- [x] The language context's `update_definition_in_context` function is updated to leverage definition UUIDs instead of names to identify previous definition versions
- [x] The language context's `add_definition_in_context` function is updated to leverage definition UUIDs instead of names to identify previous definition versions
- [x] The language context throws a LanguageError if it attempts to apply an extension to definitions with duplicate names
- [x] The language context's `remove_definition_from_context` function is updated to leverage definition UUIDs instead of names to identify previous definition versions
- [x] The language context's `get_definition_by_name` function is updated to account for the switch from definition names to UUIDS in the map key

### DIFF
--- a/python/tests/lang/test_definition.py
+++ b/python/tests/lang/test_definition.py
@@ -13,6 +13,11 @@ from tests.helpers.parsed_definitions import (
 
 
 class TestDefinition(TestCase):
+    def test_uuid(self):
+        self.assertTrue(create_schema_definition("Test").uid.is_safe)
+        self.assertEqual(create_schema_definition("Test"), create_schema_definition("Test"))
+        self.assertNotEqual(create_schema_definition("Test1"), create_schema_definition("Test2"))
+
     def test_get_fields_with_empty_no_top_level_fields(self):
         test_definition = create_schema_definition("EmptyData")
         test_definition.structure["schema"] = {}
--- a/python/src/aac/lang/definitions/definition.py
+++ b/python/src/aac/lang/definitions/definition.py
@@ -2,9 +2,7 @@
 
 import logging
 import yaml
-
 from uuid import NAMESPACE_OID, UUID, uuid5
-
 from attr import Factory, attrib, attrs, validators
 
 from aac.lang.definitions.lexeme import Lexeme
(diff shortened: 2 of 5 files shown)


python/tests/lang/test_definition.py
```diff
@@ -13,6 +13,11 @@
 
 
 class TestDefinition(TestCase):
+    def test_uuid(self):
+        self.assertTrue(create_schema_definition("Test").uid.is_safe)
+        self.assertEqual(create_schema_definition("Test"), create_schema_definition("Test"))
+        self.assertNotEqual(create_schema_definition("Test1"), create_schema_definition("Test2"))
+
```
Coffee2Bits, line 20: We could add an additional test in the definitions test to demonstrate that a large list of definitions all have unique uuids. It's not a significantly larger number than the existing test when you think about the small chance for collisions, but 60 definitions does provide a little more peace of mind. This test can go into the language context test if you'd prefer -- I was on the fence about where it should end up.

You'll also need to add the import line
```
from aac.lang.active_context_lifecycle_manager import get_active_context
```

```suggestion

    def test_uuid_in_active_context(self):
        definition_uuids_list = [definition.uid for definition in get_active_context().definitions]
        definition_uuids_set = set(definition_uuids_list)

        self.assertEqual(len(definition_uuids_list), len(definition_uuids_set))
        
```

Coffee2Bits (changes requested):
Great work! Just a minor editorial change and an additional test to demonstrate unique ids with a slightly larger set of definitions.

python/tests/lang/test_definition.py
cvchaparro: I agree, I'm going to put it in the language context tests since we're specifically accessing the active context - it seems more appropriate there, IMO.

Coffee2Bits (approved):
Looks good. Great job!

PR #440, opened 2022-09-21
LSP Rename Symbol

### LINKED ISSUE #400: I want the LSP to implement "Rename" for definition names
As an AaC user, I want the LSP to implement the "Rename" LSP function, so that I can easily rename definition and their references.

Example rename tests can be found here: https://github.com/openlawlibrary/pygls/blob/master/pygls/lsp/types/language_features/definition.py

blocked by: #439

### AC:
- [x] When a definition's name is selected and the IDE's "rename" action is invoked, then all instances of the definition's name are updated.
- [x] The LSP "prepare rename" request is implemented
- [x] The LSP "rename" request is implemented

Coffee2Bits:
Closes #400 

* Added support for the LSP Rename feature
* Added support for the LSP Prepare Rename feature

### DIFF
--- a/python/src/aac/plugins/lsp_server/providers/rename_provider.py
+++ b/python/src/aac/plugins/lsp_server/providers/rename_provider.py
@@ -56,9 +56,10 @@ class RenameProvider(lsp_provider.LspProvider):
             if symbol:
                 edits = self._get_rename_edits(symbol, new_name) if symbol else {}
                 select_symbol_text_edit = TextEdit(
-                    range=get_symbol_range_at_position(document.source, position.line, position.character)
+                    range=get_symbol_range_at_position(document.source, position.line, position.character),
+                    new_text=new_name
                 )
-                edits.get(current_uri, []).append(select_symbol_text_edit)
+                edits.get(document.path, []).append(select_symbol_text_edit)
 
                 try:
                     workspace_edit = WorkspaceEdit(text_document=TextDocumentIdentifier(uri=document.uri), changes=edits)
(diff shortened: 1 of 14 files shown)


cvchaparro (changes requested):
I'm not sure if it's just my environment but I saw a few things during my testing that seemed to not work.

Specifically, I tried renaming some things in the `alarm_clock` example project. I was successfully able to do a rename `Clock` to `Crock` when I executed the rename action [here](https://github.com/jondavid-black/AaC/blob/5596ad5cfeb497c08c085c525cd99cede3013692/python/model/alarm_clock/alarm_clock.yaml#L8). After that, I tried renaming `Crock` back to `Clock` and I got an error stating that "Rename failed to apply edits". After that, I tried changing the name of `ClockTimer` to `ClockTime` [here](https://github.com/jondavid-black/AaC/blob/5596ad5cfeb497c08c085c525cd99cede3013692/python/model/alarm_clock/alarm_clock.yaml#L39) and I got the same error (in fact, after the first rename I continually got this error message).

I also tried a few other things but the error remained. In looking at the logs, it looks like it might be an issue with gathering some of the symbols but I'm not 100% sure on that.

python/src/aac/plugins/lsp_server/providers/rename_provider.py
```diff
@@ -52,10 +52,9 @@ def get_rename_edits(
 
         if document:
             symbol = get_symbol_at_position(document.source, position.line, position.character)
+            edits = self._get_rename_edits(symbol, new_name) if symbol else {}
 
             if symbol:
```
cvchaparro, line 57: I'm pretty sure this if statement is not longer needed since `edits` will be set to a valid value depending on whether `symbol` is set. Nothing else uses symbol in the `try`/`except` block.

Coffee2Bits: The if state is necessary since the response to a null symbol is a None response rather than a WorkspaceEdit with 0 changes.

cvchaparro (approved):
Nice work!

PR #452, opened 2022-09-30
Schema Inheritance

### LINKED ISSUE #441: I want Schema definitions to support modeling Data Structure Inheritance
As an AaC user, I want Schema definitions to support modeling Data Structure Inheritance, so that I can model existing or potential data structures that leverage inheritance.

## Notes:

New Reference Schema definition:
```python
schema:
  name: DefinitionReference
  fields:
    - name: name
      type: string
  validation:
    - name: Type references exist
      arguments:
        - name
```

Updated Schema definition:
```python
schema:
  name: Field
  fields:
    - name: name
      type: string
      description: |
        The name of the field.
        Used to refer to the field, in certain cases.
    - name: type
      type: string
      description: |
        The type of the field.
        Used to constrain the field's values to make it simpler to reason about. Must
        reference an existing definition by name.
    - name: inherits
      type: DefinitionReference[]
      description: |
        A list of Schema definition names that this definition inherits from.
```

Example Schema with inheritance:
```python
schema:
  name: DataB
  inherits:
    - name: DataA
  fields:
    - name: metadata
      type: MessageMetadataData
    - name: transformed_msg
      type: string
  validation:
    - name: Required fields are present
      arguments:
        - metadata
```

## AC:
- [x] A new Schema is for definition references is added to the core spec `DefinitionReference`
- [x] Schema definitions have a new field `inherits`
  - [x] The field will reference one or more other schema definitions 
  - [x] The field will use a new AaC core data structure `DefinitionReference` to define superclass definitions
- [x] Schema definitions will inherit the following transitive attributes from their parent definition(s)
  - [x] Fields
  - [x] Validations
- [x] The `alarm_clock` model in the `src/python/model` directory is updated with an example of inheritance
- [x] There are tests demonstrating schema inheritance
  - [x] The parent definition(s) can be obtained from the sub definition
  - [x] The transitive superclass fields are present in the sub-class
  - [x] The transitive superclass validations are present in the sub-class

Coffee2Bits:
Closes #441 

* Added a new field to `schema` called named `inherits`
* Added support for modeling data structure inheritance in `schema` definitions
* Added initial support for constants pertaining to the foundational/core AaC definitions
* Audited the runtime dependencies
* Expanded the `get_<field>` API for core DSL field names
* Expanded the flow model to leverage data inheritance as an example


python/tests/plugins/lsp_server/providers/test_goto_definition_provider.py
```diff
@@ -71,7 +71,7 @@ async def test_get_named_location(self):
         location, *_ = self.provider.get_definition_location_at_position(
             {document_uri: self.virtual_document_to_lsp_document(TEST_DOCUMENT_NAME)},
             document_uri,
-            Position(line=29, character=13),
+            Position(line=25, character=13),
```
Coffee2Bits, line 74: This position changed due to some changes in the `parsed_definition` file.

Coffee2Bits:
I found the alarm clock example to be too simple to demonstrate the use for `schema` inheritance, so I will modify the flow model instead to leverage schema inheritance as an example.

python/model/flow/DataMessage.yaml
```diff
@@ -0,0 +1,12 @@
+schema:
+  name: DataMessage
+  fields:
```
Coffee2Bits, line 3:
```suggestion
  name: DataMessage
  description: Provides metadata for flow system messages.
  fields:
```

python/src/aac/lang/language_context.py
```diff
@@ -83,15 +95,40 @@ def add_definitions_to_context(self, definitions: list[Definition]):
         Args:
             definitions: The list of Definitions to add to the context.
         """
-        extension_definitions = get_definitions_by_root_key("ext", definitions)
+
+        def dirty_dependency_sort_for_definitions_with_inheritance(definitions: list[Definition]) -> list[Definition]:
```
cvchaparro, line 99: What does "dirty" mean, here? Is it a reference to the possibility of duplicated parent definitions? If so, could we eliminate that using a set?

cvchaparro (changes requested):
Nice work; I have a couple minor suggestions.

Also, I am curious if, after working this, you have any ideas on how we might turn this into a plugin? It looks like it could be done pretty easily with some hooks/callbacks/etc or by having a subclass of the language context (although, admittedly, I haven't thought too much about whether that would introduce more issues).

python/src/aac/lang/language_context.py
Coffee2Bits: Dirty as in simple-fast. This is a superficial implementation until we can actually calculate dependency order for definitions. I can rename dirty -> simple here.

Coffee2Bits:
> Nice work; I have a couple minor suggestions.
> 
> Also, I am curious if, after working this, you have any ideas on how we might turn this into a plugin? It looks like it could be done pretty easily with some hooks/callbacks/etc or by having a subclass of the language context (although, admittedly, I haven't thought too much about whether that would introduce more issues).


At the moment, we don't have the infrastructure to support this with a plugin, and being able to do so may not be possible with where I've integrated the application of inherited fields. We'd need to add another contribution point and determine what value there would be to allow users to register tasks that are perform on a Definition as is added and removed from the LC.

python/src/aac/lang/language_context.py
cvchaparro: Understood. It's fine to leave it as - I just wasn't sure in what way this was "dirty".

cvchaparro (approved):
Great job!

PR #453, opened 2022-09-30
Validate a Single Definition

### LINKED ISSUE #406: I want the Validate command to support validating single definitions
As an AaC user, I want the Validate command to support validating single definitions, so that I can specify only validating one definition in a file by its name.

## Notes:
The goal is to create a public function like https://github.com/jondavid-black/AaC/blob/3c97aea699ab8be74c6aff3dc2877a6c929f3a20/python/src/aac/validate/_validate.py#L74

You may need to add all of the non-target definitions in the file to the active_context before validating the single, targeted definition.

Example usage:
```
aac validate model/flow/System.yaml --definition_name System
```

## AC:
- [x] The `validate` command has an optional argument `definition_name`
- [x] When the definition name is specified, only that definition in the accompanying architecture file is validated.
- [x] An error message is returned if there is no definition with the target name is in the file.
  - [x] If the definition exists in the context, report its parent file
  - [x] If the definition doesn't exist in the file, report the available definitions in the target file.

Coffee2Bits:
Closes #406 

* Added support for validating one definition at a time
* Added single definition validation to `aac validate`


python/tests/cli/test_validate_command.py
```diff
@@ -23,3 +27,44 @@ def test_validate_command_fails_with_invalid_definition(self):
...
+            active_context = get_active_context()
+            self.assertTrue(active_context.is_definition_type(valid_model.name))
+
+            # This check will only work if we allow validate to pollute the active context for validation.
+            self.assertFalse(active_context.is_definition_type(valid_schema.name))
+
+    def test_validate_command_specify_definition_name_fails_with_missing_definition(self):
+        valid_model = create_model_definition("Valid Model", "A valid model.")
+        # TO/DO come back and replace this with test constant definitions
+        valid_schema = create_schema_definition("Test Schema", fields=[create_field_entry("name", "str")])
```
Coffee2Bits, line 53: This is on hold until I can integrate the new constant test definitions from #452

python/src/aac/cli/builtin_commands/validate/validate_impl.py
```diff
@@ -1,24 +1,70 @@
 """AaC Plugin implementation module for the validate plugin."""
 # NOTE: It is safe to edit this file.
 # This file is only initially generated by the aac gen-plugin, and it won't be overwritten if the file already exists.
+import logging
+from os import linesep
```
cvchaparro, line 5:
```suggestion
from os import linesep
from typing import Optional
```

python/src/aac/cli/builtin_commands/validate/validate_impl.py
```diff
@@ -1,24 +1,70 @@
...
-from aac.validate import validated_source
+from aac.validate import validated_source, validated_definition
+from aac.io.parser import parse
+from aac.lang.definition_helpers import get_definition_by_name
 
 plugin_name = "validate"
 
 
-def validate(architecture_file: str) -> PluginExecutionResult:
+def validate(architecture_file: str, definition_name: str = None) -> PluginExecutionResult:
```
cvchaparro, line 18:
```suggestion
def validate(architecture_file: str, definition_name: Optional[str] = None) -> PluginExecutionResult:
```

python/src/aac/cli/builtin_commands/validate/validate_impl.py
```diff
@@ -1,24 +1,70 @@
...
     """
     Validate the AaC definition file.
 
     Args:
         architecture_file (str): The path to the AaC file to be validated.
+        definition_name (str): The name of the definition in the file to validate.
```
cvchaparro, line 24:
```suggestion
        definition_name (Optional[str]): The name of the definition in the file to validate.
```

python/tests/cli/test_validate_command.py
```diff
@@ -23,3 +27,44 @@ def test_validate_command_fails_with_invalid_definition(self):
...
+            result = validate(valid_model_file.name, valid_model.name)
+
+            assert_plugin_success(result)
+            self.assertIn(f"{valid_model_file.name} is valid", result.get_messages_as_string())
+            self.assertIn(valid_model.name, result.get_messages_as_string())
+
+            active_context = get_active_context()
+            self.assertTrue(active_context.is_definition_type(valid_model.name))
+
+            # This check will only work if we allow validate to pollute the active context for validation.
```
cvchaparro, line 47: I suppose we still should work on deciding how to do this. Perhaps a GH discussion would be good to hash out some details?

cvchaparro (changes requested):
Nice work! Just a few simple suggestions.

python/tests/cli/test_validate_command.py
Coffee2Bits: Yeah, there are a few threads that will need to be handled when we revise validation. We could just rely on a lack of error results to indicate success assuming that we have tests demonstrating the inverse, which I believe we do.

cvchaparro (approved):

PR #409, opened 2022-10-27
I want tab-completion for the AaC package

cvchaparro:
As a user of AaC, I want tab-completion for the AaC package, so that it's easier to operate the CLI.

Note:
We're already using the click package, which supports shell completion.
https://click.palletsprojects.com/en/8.1.x/shell-completion/

## AC:
- [x] The AaC package has tab-completion in BASH
- [x] The AaC package has tab-completion in ZSH
- [x] The AaC commands have tab-completion
- [x] The AaC command arguments have tab-completion


Coffee2Bits (changes requested):
Great job! One tiny suggested edit. 

That being said, did you notice any performance delay with the tab-completion? I'm not sure if Gitpod was being slow or if we should look into speeding up the completion result.

cvchaparro:
> Great job! One tiny suggested edit. 
> 
> That being said, did you notice any performance delay with the tab-completion? I'm not sure if Gitpod was being slow or if we should look into speeding up the completion result.

On the first run it took a bit but, after that, I didn't notice much of a delay.

Coffee2Bits (approved):

PR #402, opened 2022-11-10
I want the LSP to publish validation results as diagnostics

cvchaparro:
As an AaC user, I want the LSP to publish validation results as diagnostics, so that my IDE can run validation inline and report issues as inline messages in my IDE.

# References:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics

Function used to compute the smaller dictionaries that are then validated -- this is likely the best place to calculate which tokens to provide. https://github.com/jondavid-black/AaC/blob/8ec56b5f8ca7ad65d72bd35567255af6bcbed32b/python/src/aac/lang/definitions/structure.py#L11-L73

Validation function with flag argument to validate the context as well as the target file: https://github.com/jondavid-black/AaC/blob/8ec56b5f8ca7ad65d72bd35567255af6bcbed32b/python/src/aac/validate/_validate.py#L73-L92

Root key validator is an easy initial plugin to convert since it will always be reporting feedback to the location of the first token/lexeme in the definition. https://github.com/jondavid-black/AaC/blob/8ec56b5f8ca7ad65d72bd35567255af6bcbed32b/python/src/aac/plugins/validators/root_keys/_validate_root_keys.py#L12-L35

## AC
- [x] The IDE reports validation results as LSP diagnostic findings
- [x] The LSP server implements diagnostic publishing
- [x] The LSP server validates user's AaC files when they're added to the context (didOpen)
- [x] The LSP server validates user's AaC files when they're updated in the context (didChange)
- [x] The root key validation is update to report the symbol location in the finding


python/src/aac/plugins/validators/exclusive_fields/_exclusive_fields.py
```diff
@@ -28,8 +33,17 @@ def validate_dict(dict_to_validate: dict) -> None:
...
+            multiple_exclusive_fields = (
+                f"Multiple exclusive fields are defined '{present_exclusive_fields}' in: {dict_to_validate}"
+            )
+            second_exclusive_field_lexeme = definition_under_test.get_lexeme_with_value(second)
+            findings.add_error_finding(
+                definition_under_test,
+                multiple_exclusive_fields,
+                PLUGIN_NAME,
+                *FindingLocation.from_lexeme(PLUGIN_NAME, second_exclusive_field_lexeme).to_tuple(),
+            )
```
Coffee2Bits, line 46: Future food for thought, we could alter the findings methods to accept a lexeme and do the source location conversion under the hood.

python/src/aac/plugins/validators/root_keys/_validate_root_keys.py
```diff
@@ -23,13 +28,19 @@ def validate_root_keys(definition_under_test: Definition, target_schema_definiti
     """
     findings = ValidatorFindings()
 
-    active_context_root_keys = get_active_context().get_root_keys()
+    language_context_root_keys = language_context.get_root_keys()
```
Coffee2Bits, line 31: Good catch! ❤️

python/tests/plugins/validators/test_validation_implementation_plugin.py
```diff
@@ -27,14 +31,18 @@ def test_validation_module_register_validators(self):
     def test_validate_validator_implementations_validate_validation_success(self):
         test_active_context = get_active_context(reload_context=True)
 
-        validation_definitions = list(filter(lambda definition: definition.get_root_key() == "validation", test_active_context.definitions))
+        validation_definitions = list(
+            filter(lambda definition: definition.get_root_key() == "validation", test_active_context.definitions)
+        )
```
Coffee2Bits, line 36: I'd also update to use the constants, but I won't nit pick that on a large PR like this.
```suggestion
        validation_definitions = test_active_context.get_definitions_by_root_key("validation")
```

python/tox.ini
```diff
@@ -26,13 +26,11 @@ code-directories = src
 [coverage]
 coverage = src
 always-on = True
-coverage-config = tox.ini
```
Coffee2Bits, line 29: I don't recall all the details here -- why remove this?

cvchaparro: It's a bit fuzzy because I made the change a while ago but I seem to recall it was causing an issue where the debugger was looking for the tox.ini file in the same directory as the `nose2` executable so I would never be able to debug.

Coffee2Bits (approved):
Great job! So excited to see this make its way in!

PR #455, opened 2022-11-29
I want AaC to provide persistence for the ActiveContext in the CLI.

cvchaparro:
As an AaC user, I want AaC to provide persistence for the ActiveContext in the CLI, so that I can perform long, complex manipulation of the active context from the CLI.

## References:
The active context currently recreates itself every time the CLI is invoked: https://github.com/jondavid-black/AaC/blob/2a29543e1630edd52f82f63d1f9609f9a5621551/python/src/aac/lang/active_context_lifecycle_manager.py#L10-L26

Basic documentation on reading/writing JSON files:
https://www.geeksforgeeks.org/reading-and-writing-json-to-a-file-in-python/

Filepath sanitizing function when dealing with file paths. https://github.com/jondavid-black/AaC/blob/9df2a2cf490e232753c446ecd596f398d3f6b5ff/python/src/aac/io/paths.py#L9-L34

## AC:
- [x] The Language Context provides support for exporting/importing state to/from data files
- [ ] When the active context is reloaded `get_active_context(reload_context=True)` the local persistence file is deleted
- [x] The active context lifecycle manager maintains its state via a local file
  - [x] Persistence data files are stored as JSON and serialized/deserialized using JSON
  - [x] persistence files are stored next to the AaC main file, like with logs
  - [x] On start-up, it loads the file if it is present
  - [x] On tear-down, it exports the file to the local location
  - [x] The version of the AaC package is recorded
  - [x] AaC files in the context are recorded
  - [x] Definition names in the context are recorded
  - [x] Plugins in the context are recorded
- [ ] When the active context loads in previous state, it uses the recorded AaC file names to reload definitions into the context
  - [x] AaC files in the persistent context are parsed and loaded into the active context
  - [ ] Persistent state must account for changes made to the AaC files between CLI calls


python/src/aac/plugins/plugin_execution.py
```diff
@@ -112,11 +112,6 @@ def plugin_result(name: str, cmd: Callable[..., Any], *args: Tuple[Any], **kwarg
...
-        A(n) {error_trace.name} error occurred
-          in {error_trace.filename}
-          on line {error_trace.lineno}
-        The error was: {error}
-        """
-
-    return [error_message_lines(error_trace) for error_trace in extract_tb(error.__traceback__)]
+        return f"An error occurred in {error_trace.name}\n  in file {error_trace.filename}\n  on line {error_trace.lineno}\n"
+
+    return [error_message_lines(error_trace) for error_trace in extract_tb(error.__traceback__)] + [f"The error was: {error}"]
```
cvchaparro, line 117: I cleaned up error reporting to fix extra indentation, caused by the formatted multiline string. We were also printing the same error after each block which is confusing, so I printed the error at the very end of the trace.

python/src/aac/lang/active_context_lifecycle_manager.py
```diff
@@ -21,11 +30,29 @@ def get_active_context(reload_context: bool = False) -> LanguageContext:
     global ACTIVE_CONTEXT
 
     if not ACTIVE_CONTEXT or reload_context:
-        ACTIVE_CONTEXT = get_initialized_language_context()
+        ACTIVE_CONTEXT = load_language_context(__state_file_name__, reload_context)
 
     return ACTIVE_CONTEXT
 
 
+def load_language_context(file_name: str, reload_context: bool = False) -> LanguageContext:
```
Coffee2Bits, line 38: Since this function is in the active context lifecycle manager, we should be more explicit as to which context is being loaded.
```suggestion
def load_active_context(file_name: str, reload_context: bool = False) -> LanguageContext:
```

That being said, this description implies it's more general, which would suggest it should belong as a method on the language context, which would have the user follow a pattern like:
```python
my_context = LanguageContext()
my_context.load_context_from_file(file_name)
previous_state_definitions = my_context.definitions
```

python/src/aac/lang/language_context.py
```diff
@@ -419,6 +437,17 @@ def add_plugins(self, plugins: list[Plugin]):
...
+        """
+        Add the plugin from the provided file URI.
+
+        Args:
+            names (list[str]): The names of the plugins which should be loaded. If not provided, all
+                plugins will be loaded.
+        """
+        plugins = [plugin for plugin in plugin_manager.get_plugins() if plugin.name in names]
+        self.add_plugins(plugins)
+
```
Coffee2Bits, line 450: This will need to be de-conflicted with https://github.com/jondavid-black/AaC/pull/499/files#diff-d52e144e2a86781dcfbc2e77b4ee4f8da334d641b1f37713014eec5c9a729f67R427-R430

python/src/aac/persistence/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""A package dealing with AaC persistence."""
+
+import os
+
+from aac.persistence.load import get_language_context_from_state_file
+
+
+__state_file_name__ = os.path.join(os.path.dirname(__file__), "..", "state.json")
```
Coffee2Bits, line 8: Let's make this a standard python constant name. Something like:
```suggestion
ACTIVE_CONTEXT_STATE_FILE_NAME = os.path.join(os.path.dirname(__file__), "..", "state.json")
```

python/src/aac/persistence/load.py
```diff
@@ -0,0 +1,57 @@
+"""Module for loading data from the state file and making it available to the rest of the AaC package."""
```
Coffee2Bits, line 1: I like the idea of having a module that's dedicated to persistence for other persistence concerns like user configuration, but these functions seem wholly concerned with the LanguageContext. I've been thinking about giving the LanguageContext a directory module structure into which we could drop these functions into a module like `aac.lang.language_context.persistent_state`.

python/src/aac/persistence/load.py
```diff
@@ -0,0 +1,57 @@
...
+    Returns:
+        The language context after loading definitions from the files stored the state file.
+    """
+    language_context = LanguageContext()
+    version, files, definitions, plugins = decode_state_file(file_name)
+
+    if version != __version__:
+        raise StateFileError(
+            f"Version mismatch: State file written using version {version}; current AaC version {__version__}"
+        )
```
Coffee2Bits, line 27: ❤️ This is a great idea! I wonder if we could expand on this to provide "support" version traversal (e.g. 0.1.4 -> 0.1.5 is allowed, but 0.1.4 -> 0.2.0 isn't.) That seems like a lot of overhead for current user levels; maybe in the future.

Coffee2Bits (changes requested):
Good work! I have some comments for some structural changes. I'm fine with the uncompleted AC -- they're oversights.

python/src/aac/lang/active_context_lifecycle_manager.py
```diff
@@ -21,11 +30,29 @@ def get_active_context(reload_context: bool = False) -> LanguageContext:
...
+        reload_context (bool): If True, a fresh active context will be returned; otherwise, the
+            active context will be loaded from the provided file URI if it exists.
+
+    Returns:
+        A language context object loaded from the file, if it exists; otherwise, None.
+    """
+    if lexists(file_name) and not reload_context:
+        return get_language_context_from_state_file(file_name)
+    else:
+        return get_initialized_language_context()
```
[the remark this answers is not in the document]
cvchaparro: I had this originally but I changed it because we still want to return a valid language context if the state file doesn't exist so either way `load_language_context` would return `get_initialized_language_context()` if the file doesn't exist.

To me, the approach I settled on seemed a bit cleaner to account for both cases.

python/src/aac/lang/active_context_lifecycle_manager.py
cvchaparro: Ah, good point. I'll see about moving it into the `LanguageContext` class.

python/src/aac/lang/language_context.py
cvchaparro: Yeah, I noticed that yesterday.

python/src/aac/lang/active_context_lifecycle_manager.py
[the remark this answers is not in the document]
cvchaparro: After discussing it, I'll work on moving the `import`/`export` functionality to the `LanguageContext` class. Doing it this way, would circumvent the entire issue.

cvchaparro:
I think the only thing that should be left is to ensure third-party plugins are re-registered. Currently, a third-party plugin doesn't appear in the active plugins list even though it's installed so we don't write it to the state file for reloading it on the next start up.

.github/workflows/python-tests.yml
```diff
@@ -57,9 +57,10 @@ jobs:
     - name: Test Gen-Plugin Third-Party Plugin
       run: |
         echo y | aac gen-plugin model/plugin/plugin.yaml
+        rm src/aac/state.json
```
cvchaparro, line 60: The state file left over from the last pipeline stage was causing the test plugin to not be available so removing it here should allow starting fresh to import all definitions and activate all plugins.

Coffee2Bits (approved):
Great job!

PR #498, opened 2022-11-29
Adding Validation to the Extension type field

Coffee2Bits:
* Added `Type references exist` to `extension` so that the field `type` is a valid definition reference
* Fixed type hint


Coffee2Bits:
This validation was missing, so users weren't seeing any error indications when writing extensions that target undefined types -- which is now validated with this PR.

cvchaparro (changes requested):
When I test this, I'm getting an error in `heirarchy.py` that I think can be resolved by changing this line:
https://github.com/jondavid-black/AaC/blob/06411ca10263b9350acdd250e8239154a16cb3aa/python/src/aac/lang/hierarchy.py#L25

to this:
```python
    while not found_self_defined_ancestor and ancestor_definition:
```

python/src/aac/plugins/validators/defined_references/_validate_references.py
```diff
@@ -37,15 +39,26 @@ def validate_dict(dict_to_validate: dict) -> None:
...
+                        logging.debug(f"Value '{field_reference}' doesn't exist in definition. Likely an extension value.")
+
             else:
                 missing_field_in_dictionary = f"Missing field 'type' in validation content dictionary: {dict_to_validate}"
+                logging.debug(missing_field_in_dictionary)
                 name_lexeme = definition_under_test.get_lexeme_with_value(definition_under_test.name)
+
+                if not name_lexeme:
+                    name_lexeme, *_ = definition_under_test.lexemes
+                    name_lexeme = name_lexeme or Lexeme(SourceLocation(0, 0, 0, 0), definition_under_test.source.uri, definition_under_test.name)
```
cvchaparro, line 58: Was a missing value for `name_lexeme` causing the error mentioned in my previous review? I'm wondering if there will be unexpected ramifications of using a default `SourceLocation` of the first position in the file.

cvchaparro (approved):

PR #503, opened 2022-12-06
Validation - Boolean Primitive Type

### LINKED ISSUE #487: I want a primitive boolean validation
As an AaC user, I want a primitive validation for booleans, so that I can guarantee that fields with booleans values are represented as "true" or "false".

Partially address #329 

## References:

YAML Specification support for boolean values: https://stackoverflow.com/a/53648738

## AC:
- [x] There is a new primitive type validation for booleans
- [x] If the boolean values are treated as strings:
  - [x] The boolean primitive validation ensures that instances of "boolean" are true/false string values.
  - [x] When the values are compared, the string values should be converted to lowercase then compared -- allowing the user to use case insensitive FALSE/true values.
- [x] If a user puts a non-boolean value in a boolean field, their validation fails with an appropriate error message.

JustAChucklin:
closes #487 
- [x] There is a new primitive type validation for booleans
- [x] If the boolean values are treated as strings:
  - [x] The boolean primitive validation ensures that instances of "boolean" are true/false string values.
  - [x] When the values are compared, the string values should be converted to lowercase then compared -- allowing the user to use case insensitive FALSE/true values.
- [x] If a user puts a non-boolean value in a boolean field, their validation fails with an appropriate error message.


python/src/aac/plugins/first_party/primitive_type_check/validators/bool_validator.py
```diff
@@ -0,0 +1,39 @@
...
+def get_validator() -> PrimitiveValidationContribution:
+    """Return the Primitive Validator for 'boolean'."""
+    return PrimitiveValidationContribution(BOOL_VALIDATION_NAME, PRIMITIVE_TYPE_BOOL, validate_bool)
+
+
+def validate_bool(definition: Definition, value_to_validate: Any) -> Optional[ValidatorFinding]:
+    """
+    Returns a Validation finding if the type isn't valid, otherwise None.
+
+    `This function is intended to be used in the Validation apparatus, and for this result
```
Coffee2Bits, line 20:
```suggestion
    This function is intended to be used in the Validation apparatus, and for this result
```

python/src/aac/plugins/first_party/primitive_type_check/validators/bool_validator.py
```diff
@@ -0,0 +1,39 @@
...
+    to be collated with other finding into a larger ValidatorResult.
+
+    Arge:
+        definition (Definition): The definition that the value belongs to.
+        value_to_validate (Any): The value to be tested.
+
+    Returns:
+        `A ValidatorFinding if the value is not a boolean or None.
```
Coffee2Bits, line 28:
```suggestion
        A ValidatorFinding if the value is not a boolean or None.
```

Coffee2Bits (changes requested):
Looks good! Just need to register the bool validation to the plugin in order for the validate function to pick it up.

You'll need:
```python
def _get_primitive_validations() -> list[PrimitiveValidationContribution]:
    return [
        int_validator.get_validator(),
        bool_validator.get_validator(),
    ]
```
Here: https://github.com/jondavid-black/AaC/blob/464c7909e05a8f8dfec74705efa6a8a9b078d234/python/src/aac/plugins/first_party/primitive_type_check/__init__.py#L23-L26

After that, you should be able to force the error with the validate CLI command on a bad model.

JustAChucklin:
Ok pushed the commit through...I think your comments have been addressed @Coffee2Bits

PR #504, opened 2022-12-07
LanguageContext Update AaC File

### LINKED ISSUE #364: The LanguageContext should support updating the definition content in files
We currently don't have a method to propagate updates in the LanguageContext to the definition's source file, which is an issue for features that update definitions based on changes NOT made in the definition's source file. (e.g. the LSP vs the REST API.) 

## AC:
- [ ] The LanguageContext provides a new feature to synchronize the definitions in the context with their source files.
  - [ ] Definitions present in the file, but not the Language Context are removed from the file
  - [ ] Definitions in the file that differ from their LanguageContext counterparts are overwritten (source of truth is the LanguageContext)

Coffee2Bits:
Closes #364

* LanguageContext now supports pushing updated definitions to their respective AaC file
* Updating definitions via the REST API will result in updated AaC files
* Deleting a definition via the REST API removes it from the file
* Delete the last definition in a file via the REST API will remove the file


cvchaparro (changes requested):
Here are a few minor suggestions but, overall, I am still seeing a complete overwrite of a file when adding a new definition.

Coffee2Bits:
> Here are a few minor suggestions but, overall, I am still seeing a complete overwrite of a file when adding a new definition.

I'm not able to recreate this behavior.
Here's a screencap of the REST-API adding a new definition onto the end of the file. Can you provide details to recreate?

PR #508, opened 2022-12-14
Hotfix for some issues with plugin activation

Coffee2Bits:
* Fixed an exception when activating/deactivating a plugin by an invalid name


cvchaparro (approved):
Good catch!

PR #509, opened 2022-12-16
LanguageContext Prevents changes to Non-editable files

### LINKED ISSUE #365: The LanguageContext must only allow updates to definitions sourced from user-defined files
The LanguageContext must only allow updates to definitions sourced from user-defined files. The goal is to prevent users from changing definitions sourced from the core spec, plugins, and 3rd party user libraries. 

## AC:
- [x] If a definition's AaCFile's attribute `is_user_editable` is true, then the LanguageContext will allow the definition to be updated
- [x] If a definition's AaCFile's attribute `is_user_editable` is false, then the LanguageContext will not allow the definition to be updated and will log a warning.

Coffee2Bits:
Closes #365 

* The context now prohibits updates to definitions in non-editable files
* The context now prohibits removing definitions from non-editable files
* Non-editable files can still be added to the context


python/src/aac/lang/language_context.py
```diff
@@ -173,7 +176,7 @@ def remove_definition_from_context(self, definition: Definition):
         Args:
             definition (Definition): The Definition to remove from the context.
         """
-        if definition.uid in self.definitions_dictionary:
+        if definition.uid in self.definitions_dictionary and self._is_definition_editable(definition):
```
cvchaparro, line 179: I think we should make the `self._is_definition_editable(...)` call another `if` statement so we can give an error message that gives the user an appropriate understanding of what happened. Namely, if I try to delete `Primitives` I shouldn't be told that the definition isn't present but rather that the definition cannot be edited.

python/src/aac/lang/language_context.py
```diff
@@ -219,12 +222,11 @@ def update_definition_in_context(self, definition: Definition):
             definition (Definition): The Definition to update in the context.
         """
 
-        if definition.uid in self.definitions_dictionary:
-            old_definition = self.definitions_dictionary.get(definition.uid)
-
+        old_definition = self.definitions_dictionary.get(definition.uid)
+        if old_definition and self._is_definition_editable(definition):
```
cvchaparro, line 226: Similar to my comment above, we should also give a specific error message here for the case when someone tries to update a non-editable definition (i.e. `Primitives`, etc).

python/src/aac/lang/language_context.py
```diff
@@ -219,12 +222,11 @@ def update_definition_in_context(self, definition: Definition):
...
             self.remove_definition_from_context(old_definition)
             self.add_definition_to_context(definition)
-        else:
+        elif definition.uid not in self.definitions_dictionary:
             definitions_in_context = self.get_defined_types()
             logging.error(
                 f"Definition not present in context, can't be updated. '{definition.name}' not in '{definitions_in_context}'"
```
cvchaparro, line 232: I can't leave a comment on the other lines but this also applies to lines 185-187. I definitely think we should log an error but (at least for the Rest API) only logging an error message doesn't give the user feedback that what they tried to do failed unless they happen to look at the logs (but I wouldn't assume someone would do so when the API indicates that it succeeded). This is probably something we need to consider making a bug to handle later because actually raising an error may cause issues elsewhere, but I think we should be aware of the current limitation.

cvchaparro (changes requested):
There are a few comments on this one - let me know if you have any questions about it.

python/src/aac/lang/definitions/definition.py
```diff
@@ -82,12 +90,14 @@ def get_top_level_fields(self) -> dict[str, dict]:
...
-        return fields.get(DEFINITION_FIELD_DESCRIPTION)
+        description = fields.get(DEFINITION_FIELD_DESCRIPTION)
+        return str(description) if description else None
 
     def get_type(self) -> Optional[str]:
         """Return the string for the extension type field, or None if the field isn't defined."""
         fields = self.get_top_level_fields()
-        return fields.get(DEFINITION_FIELD_TYPE)
+        type_field = fields.get(DEFINITION_FIELD_TYPE)
+        return str(type_field) if type_field else None
```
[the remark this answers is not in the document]
Coffee2Bits: This is to enforce that the returned type is actually [str, None] since a dict can return a non-string value and downstream functions may assume non-None is a str and therefore capable of performing string methods which obviously throw errors if the object returned from the dict isn't a string.

python/src/aac/lang/language_context.py
Coffee2Bits: This should be covered by the lang error raised when this happens.

python/src/aac/lang/language_context.py
Coffee2Bits: Agreed & implemented

cvchaparro (changes requested):
A few more minor changes and it should be good to go.

cvchaparro (approved):
Nice work!

PR #445, opened 2022-12-20
I want CLI commands to manage AaC Plugins

cvchaparro:
As an AaC user, I want CLI commands to manage AaC Plugins, so that I can choose which plugins are active, inactive, and manage them with a finer control than leveraging pip to install and uninstall packages.

Blocked by:
* #455
* #456

## AC:
- [x] There is a new first-party plugin called `plugin-management`
- [x] The new plugin has a command to view all installed plugins `aac list-plugins --all`
- [x] The new plugin has a command to view the currently active plugins `aac list-plugins --active`
  - [x] The plugin lists all of the plugin names and descriptions for the plugins in the active context.
- [x] The new plugin has a command to view inactive plugins `aac list-plugins --inactive`
  - [x] The plugin lists all of the plugin names and descriptions for the plugins installed, but not in the active context.
- [x] The new plugin has a command to activate an inactive plugin `aac activate-plugin <plugin_name>`
  - [x] An error message is returned if the plugin can't be found, is already active, or can't be activated
- [x] The new plugin has a command to deactivate an active plugin `aac deactivate-plugin <plugin_name>`
  - [x] An error message is returned if the plugin can't be found, is already inactive, or can't be deactivated


cvchaparro:
I have the `list-plugins` command implemented but there's one issue I've run
into with it. Specifically, `click` is currently requiring an argument. I can
specify that each flag has `number_of_arguments=0` but that causes another
eccentricity with click that I need to investigate.

python/src/aac/plugins/first_party/plugin_management/plugin-management.yaml
```diff
@@ -0,0 +1,113 @@
...
+      input:
+        - name: plugin-name
+          type: string
+          python_type: str
+          description: The name of the plugin to be activated.
+      acceptance:
+        - scenario: Plugin is activated.
+          given:
+            - The plugin named {{plugin-management.activate-plugin.plugin-name}} is installed on the system.
+            - The plugin named {{plugin-management.activate-plugin.plugin-name}} is inactive on the system.
```
Coffee2Bits, line 64: It sounds like we'd want another plugin for this scenario since only the `plugin-management` plugin can install plugins.
```suggestion
            - The plugin named {{test-plugin.plugin-name}} is installed on the system.
            - The plugin named {{test-plugin.plugin-name}} is inactive on the system.
```

python/src/aac/plugins/first_party/plugin_management/plugin-management.yaml
```diff
@@ -0,0 +1,113 @@
...
+      input:
+        - name: plugin-name
+          type: string
+          python_type: str
+          description: The name of the plugin to be deactivated.
+      acceptance:
+        - scenario: Plugin is deactivated.
+          given:
+            - The plugin named {{plugin-management.activate-plugin.plugin-name}} is installed on the system.
+            - The plugin named {{plugin-management.activate-plugin.plugin-name}} is active on the system.
```
Coffee2Bits, line 95: Same comment as above.

Coffee2Bits (changes requested):
Great job!! I left a couple suggestions for the plugin yaml file. 

I did notice that running `aac list-plugins` returns  the output `No option provided`. How difficult would it be to have the command return some help output indicating the available flags instead of `No option provided`? It looks like it's possible for `_get_selected_option`, but I suspect it would need some additional args/info.

python/src/aac/plugins/first_party/plugin_management/plugin-management.yaml
cvchaparro: Resolving per our discussion.

python/src/aac/plugins/first_party/plugin_management/plugin-management.yaml
cvchaparro: Resolving per our discussion.

python/src/aac/plugins/first_party/plugin_management/plugin-management.yaml
```diff
@@ -0,0 +1,113 @@
...
+        - scenario: Error when multiple flags given.
+          when:
+            - More than one flag is provided.
+          then:
+            - An error is displayed to the user stating that flags are mutually exclusive.
+    - name: activate-plugin
+      type: command
+      description: Activate a plugin that's available and inactive on the system.
+      input:
+        - name: plugin-name
```
cvchaparro, line 56:
```suggestion
        - name: name
```

python/src/aac/plugins/first_party/plugin_management/plugin-management.yaml
```diff
@@ -0,0 +1,113 @@
...
+            - The plugin named {{plugin-management.activate-plugin.plugin-name}} is active.
+          when:
+            - The command is run with {{plugin-management.activate-plugin.plugin-name}}.
+          then:
+            - An error message is displayed to the user stating the plugin is active.
+    - name: deactivate-plugin
+      type: command
+      description: Deactivate a plugin that's available and active on the system.
+      input:
+        - name: plugin-name
```
cvchaparro, line 87:
```suggestion
        - name: name
```

cvchaparro (commented):
Match input names to parameter names (this is needed because there was a clash between the global `plugin_name` variable and the parameters also called `plugin_name`).

cvchaparro:
> Great job!! I left a couple suggestions for the plugin yaml file.
> 
> I did notice that running `aac list-plugins` returns the output `No option provided`. How difficult would it be to have the command return some help output indicating the available flags instead of `No option provided`? It looks like it's possible for `_get_selected_option`, but I suspect it would need some additional args/info.

That was [this](https://github.com/jondavid-black/AaC/blob/467853bae55b11fa58a2477f6a108cbe29dd94d6/python/src/aac/cli/execute.py#L65) is supposed to do for us but for some reason it wasn't working for this command. I can reinvestigate it, though.

Coffee2Bits (approved):

PR #512, opened 2023-01-11
LSP Rename Root Keys

### LINKED ISSUE #448: I want the LSP Rename to support Root Keys
As an AaC user, I want the LSP Rename to support Root Keys, so that I can rename all instances of root keys.

## AC:
- [ ] When rename is performed on a root key symbol, all instances of the root key are updated
- [ ] When rename is performed on a root key symbol, the `root` definition name is updated

Coffee2Bits:
Closes #448

* Changed runtime versions to [compatible releases](https://peps.python.org/pep-0440/#compatible-release)
* Added the LSP Rename support for root keys


Coffee2Bits:
Turns out if the LSP server throws an exception it’s not clear via the test apparatus — I suspect this is because of the async nature of the test framework. The failures manifest in the logs and some of the language server attributes being populated with debug traces during debugging.

I had to add some null checks in the LSP server to get around some of issues the startup sever not having the LSP Workspace populated when some of the LSP providers, like diagnostics, tried to access the Workspace object despite it being `None`.

python/setup.py
```diff
@@ -12,24 +12,25 @@
     readme_description = ""
 
 runtime_dependencies = [
```
cvchaparro, line 14: Future request but maybe we can do this kind of change in its own PR to limit testing dependency issues versus PR issues.

python/src/aac/plugins/first_party/lsp_server/language_server.py
```diff
@@ -188,63 +188,50 @@ async def did_change(ls: AacLanguageServer, params: DidChangeTextDocumentParams)
...
+
+def _provider_handle_request(provider_name: str, ls: AacLanguageServer, params: Any):
+    provider = ls.providers.get(provider_name)
+
+    if provider:
+        diagnostics_results = provider.handle_request(ls, params)
+        logging.debug(f"Publish Diagnostics results for '{provider_name}': {diagnostics_results}")
+        return diagnostics_results
+    else:
+        logging.warning(f"Unable to publish diagnostics; no provider found for '{provider_name}'.")
```
cvchaparro, line 237: This isn't really related to diagnostics - each provider had their own results so, to better match previous functionality, I think we should do it like this:

```suggestion
        results = provider.handle_request(ls, params)
        logging.debug(f"{provider_name} results: {results}")
        return results
    else:
        logging.warning(f"No provider found for '{provider_name}'.")
```

python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
```diff
@@ -87,6 +87,16 @@ def _get_rename_edits(self, symbol: str, new_name: str) -> dict[str, TextEdit]:
             else:
                 edits = self._get_enum_value_type_text_edits(name, new_name, enum_to_find, language_context)
 
+        if SymbolType.ROOT_KEY in symbol_types:
+            root_fields = language_context.get_root_fields()
+            key_schema_field, *_ = [field for field in root_fields if field.get("name") == name]
+            definition_to_find = language_context.get_definition_by_name(key_schema_field.get("type"))
```
cvchaparro, line 93: You'd need to add the imports, but we should use our constants:

```suggestion
            key_schema_field, *_ = [field for field in root_fields if field.get(DEFINITION_FIELD_NAME) == name]
            definition_to_find = language_context.get_definition_by_name(key_schema_field.get(DEFINITION_FIELD_TYPE))
```

python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
```diff
@@ -87,6 +87,16 @@ def _get_rename_edits(self, symbol: str, new_name: str) -> dict[str, TextEdit]:
...
+
+            if not definition_to_find:
+                logging.critical(f"Can't find the source definition definition '{name}'.")
```
cvchaparro, line 96:
```suggestion
                logging.critical(f"Can't find the source definition '{name}'.")
```

python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
```diff
@@ -120,7 +130,27 @@ def _get_enum_value_type_text_edits(
...
         enum_references_to_alter = [*enum_references, definition_to_find]
 
-        for definition in enum_references_to_alter:
+        for definition in _filter_editable_definitions(enum_references_to_alter):
+            reference_lexemes = [lexeme for lexeme in definition.lexemes if remove_list_type_indicator(lexeme.value) == old_value]
+            for lexeme_to_replace in reference_lexemes:
+                document_edits = edits.get(lexeme_to_replace.source, [])
+                replacement_range = get_location_from_lexeme(lexeme_to_replace).range
+                document_edits.append(TextEdit(range=replacement_range, new_text=new_value))
+                edits[str(lexeme_to_replace.source)] = document_edits
```
cvchaparro, line 139: We should probably turn this into a separate function and re-use it in both `_get_*_text_edits` functions.

cvchaparro (changes requested):

python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
Coffee2Bits: Embarrasing that I didn't do that.

python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
Coffee2Bits: Done

cvchaparro (approved):
Nice work!

PR #476, opened 2023-01-17
I want a plugin that can manage the active context

cvchaparro:
As an AaC CLI user, I want a plugin that can manage the active context, so that I can manage my active context from the CLI.

* Blocked by #455

* Added the active-context plugin for managing the context via the CLI
 * Users can list the files in the active context
 * Users can remove files from the active context
 * Users can add files to the active context
 * Users can reset the context
 * Users can list the definitions in the context
 * Users can describe specific definitions
 * Users can import state from another context
 * Users can export state from the active context

## AC:
- [x] There is a new plugin `active-context`
- [x] The new plugin allows users to do the following with the active context
  - [x] Add an AaC file to the context
  - [x] Remove an AaC file from the context
  - [x] List all of the files in the context
  - [x] List all of the definitions in the context
  - [x] Describe a definition by name
    - [x] Provide the structure of the definition corresponding to the provided name (to_yaml)
    - [x] Provide the definition's source file and starting line
    - [x] If there is no definition with the provided name, return an error stating as much
  - [x] Reset the context (remove user-provided definitions and files. Leaves active plugins.)
  - [x] Import state
  - [x] Export state


python/src/aac/plugins/first_party/active_context/active_context_impl.py
```diff
@@ -0,0 +1,161 @@
...
+    Args:
+        state_file (str): The persistent state file from which to get information about how to configure the active context.
+    """
+
+    def import_state_file_if_present() -> str:
+        if not lexists(state_file):
+            raise PluginError(f"The state file {state_file} could not be imported because it doesn't exist.")
+
+        active_context = get_active_context()
+        active_context.clear()
```
cvchaparro, line 133: Currently, the quickest way I found to avoid a duplicate definition error when importing was to do this. I'd still need to debug the issue to see why this is cropping up and how to resolve it.

Coffee2Bits: We should be clearing the state when importing files anyway, I've moved this clear call to the LC's `import_from_file` method.

python/src/aac/cli/execute.py
```diff
@@ -18,8 +18,8 @@ def cli():
 @cli.result_callback()
 def output_result(result: PluginExecutionResult):
     """Output the message from the result of executing the CLI command."""
-    error_occurred = not result.is_success()
-    echo(result.get_messages_as_string(), err=error_occurred, color=True)
+    error_occurred = not result.is_success
```
Coffee2Bits, line 21:
```suggestion
    error_occurred = not result.is_success()
```

Coffee2Bits (approved):

PR #518, opened 2023-01-24
Validator Error Handling

### LINKED ISSUE #341: Validator plugins that experience exceptions should not prevent the remaining validator plugins from executing normally
Validator plugins that experience exceptions should not prevent the remaining validator plugins from executing normally.

## References:
Definition validation function: https://github.com/jondavid-black/AaC/blob/3dfc11baf0ec8d7250608bd16caebdae837fc8af/python/src/aac/validate/_validate.py#L136-L150

Primitive validation function: https://github.com/jondavid-black/AaC/blob/3dfc11baf0ec8d7250608bd16caebdae837fc8af/python/src/aac/validate/_validate.py#L193-L206

## AC:
- [ ] Validator plugins are executed with error-handling that doesn't prevent the execution of other validation plugins if one of them raises an exception.
- [ ] Definition validation has a try/except error handler implemented. 
  - [ ] The exception is logged
  - [ ] The exception is returned to the user as an error message
- [ ] Exceptions that occur during validation are treated as if they were errors reported by the plugins
  - [ ] Definition validation
  - [ ] Primitive validation
- [ ] There is a test that demonstrates a validation plugin throwing a runtime exception, but not preventing results from other validation plugins.

Coffee2Bits:
Closes #341

* Added error handling for validation
* Errors encountered by validators will be reported as validation failures


jondavid-black (approved):
Improvements to validators will be nice.

PR #521, opened 2023-01-25
extended model to include requirements list

jondavid-black:
Very simple mod to allow requirements to be traced to a model.  Preserved the ability to trace requirements to a specific behavior on a model.

### DIFF
--- a/python/src/aac/plugins/first_party/specifications/specifications.yaml
+++ b/python/src/aac/plugins/first_party/specifications/specifications.yaml
@@ -36,7 +36,7 @@ ext:
 ---
 ext:
   name: addSpecTraceToModel
-  type: Model
+  type: model
   schemaExt:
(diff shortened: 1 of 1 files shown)


python/src/aac/plugins/first_party/specifications/specifications.yaml
```diff
@@ -34,6 +34,14 @@ ext:
       - name: spec
         type: Specification
 ---
+ext:
+  name: addSpecTraceToModel
+  type: Model
```
Coffee2Bits, line 39: This fixed it for me. This is something I'd like to address by developing an AaC coding style and then standardizing the core spec and plugins.
```suggestion
  type: model
```

Coffee2Bits (approved):
Looks good to me. 👍

PR #522, opened 2023-01-25
Disabling persistence

Coffee2Bits:
Disabling persistence for performance reasons.


jondavid-black (approved):
Nice and simple change!  Thanks!

PR #523, opened 2023-01-27
Duplicate Import Bugfix

### LINKED ISSUE #517: Specification validation for unique names
When importing both a "parent" and a "grandparent" specification where the "parent" also imports the "grandparent", validation results in an invalid error ```Validation finding from 'Unique definition names' of level ERROR```.  It appears to "double validate" on each import resulting in the error.  This can be resolved by knowing about the import of the "grandparent" within the "parent" and omitting the "grandparent" import.

Desired behavior:  do not "double check" for unique names between direct and indirect imports.

I consider this a bug because:
1. The user may not comprehend all the nested imports that may be present.
2. If you reference requirements from multiple spec files, the user should have a reasonable expectation that each spec file should be imported without resulting in an error.

Example:
grandparent.yaml
```yaml
spec:
  name: Grandparent Specification
  description:  A high level set of requirements.
  sections:
    - name: A Great Section
      description: "A set of great high level requirements."
      requirements:
        - id: System_01
          shall: "The system shall be awesome."
        - id: System_02
          shall: "The system shall do cool stuff."
```
parent.yaml
```yaml
import:
  - grandparent.yaml
spec:
  name: Parent Specification
  description:  A refined set of requirements.
  sections:
    - name: A More Detailed Section
      description: "A set of great more detailed requirements."
      requirements:
        - id: Module_01
          shall: "The system shall be super cool."
          parent:
            ids:
              - "System_01"
```
child.yaml
```yaml
import:
  - grandparent.yaml
  - parent.yaml
spec:
  name: Child Specification
  description:  A low level set of requirements.
  sections:
    - name: A Super Detailed Section
      description: "A set of great super detailed requirements."
      requirements:
        - id: Component_01
          shall: "The system shall be totally bodacious."
          parent:
            ids:
              - "System_02"
              - "Module_01"
```

Running ```aac validate child.yaml``` should result in the 'Unique definition names' error although everything is named uniquely.  Removing the import of 'grandparent.yaml' and rerunning ```aac validate child.yaml``` should result in no error.

Coffee2Bits:
Fixes #517 

* Fixed duplicate import bug that caused the same file to be loaded into the context multiple times if the relative paths were equivalent but not equal.

### DIFF
--- a/python/model/alarm_clock/alarm_clock.yaml
+++ b/python/model/alarm_clock/alarm_clock.yaml
@@ -24,6 +24,8 @@ model:
           then:
             - The alarm is triggered when the clock reaches the specified time
 ---
+import:
+  - ../alarm_clock/structures.yaml
 model:
   name: Clock
   description: A simple clock that keeps track of the current time
--- a/python/src/aac/io/parser/_parse_source.py
+++ b/python/src/aac/io/parser/_parse_source.py
@@ -217,6 +217,7 @@ def _get_files_to_process(arch_file_path: str) -> list[str]:
         for imp in root["import"]:
             arch_file_dir = path.dirname(path.realpath(arch_file_path))
             parse_path = path.join(arch_file_dir, imp.removeprefix(f".{path.sep}"))
-            files_to_import.update(_get_files_to_process(parse_path))
+            sanitized_path = sanitize_filesystem_path(parse_path)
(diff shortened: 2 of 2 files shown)


jondavid-black (approved):
Glad there was a simple fix.

PR #525, opened 2023-01-27
CLI errors don't return error codes

Coffee2Bits:
I found that the error code for failed CLI commands wasn't being propagated to the CLI -- looks like a small error with integrating Click.

* Fixed an issue preventing the return of error codes from commands


jondavid-black (approved):
That one looked a little difficult.  Thanks for pushing through it all.

Issue #329, opened 2022-05-10
Validate primitives


jondavid-black:
As a modeler, I want validation to ensure my primitive types are valid so that I get immediate feedback on errors in my model content.

AC:

- [x] Must ensure types identified as int are in fact integer values.
- [x] Must ensure types identified as numbers are in fact decimal values
- [x] Must ensure types identified as boolean are in fact true or false values
- [x] Must ensure types identified as date do in fact parse as date values values
- [x] Must ensure types identified as file are in fact existing files

Note:  Not sure if/how we would validate the map type.  Is the map type even needed or used?  Also not sure there's any real validation to be done for strings other than maybe ensuring it is not empty.

Coffee2Bits:
Considering this addressed by the linked PRs.

PR #489, opened 2023-01-28
I want a primitive file validation

cvchaparro:
As an AaC user, I want a primitive validation for files, so that I can guarantee that fields with file values are valid file paths on the machine.

Partially address #329 

## References:

Example to check if a path is a file `os.path.isfile(path)`: https://linuxize.com/post/python-check-if-file-exists/

## AC:
- [x] There is a new primitive type validation for files
- [x] The file primitive validation ensures that instances of "file" are valid files that exist on the machine. 
- [x] If a user puts a non-file path value or a file path that doesn't exist into a file primitive field, their validation passes with an appropriate warning message.


Coffee2Bits (approved):
Looks good!

PR #527, opened 2023-01-29
526 refactor specification plugin

jondavid-black:
This PR updates the specification plugin to add validators for globally unique requirement ids and ensuring any requirement id references are valid.  It also changes the old command intended to do this validation to a csv report generator.  Additionally it enables engineers to trace requirements bi-directionally for parent and child relationships.

It also fixes a bug discovered during testing.  There was an issue where "circular imports" would "parse forever" and then crash aac.  The parser will not ensure the same file is not parsed more than once.


jondavid-black:
Draft Release Notes:

Breaking Changes:
  - The spec-validate command no longer exists.

Features:
  - The spec-csv command has been added to output "excel ready" csv files from requirement specifications.
  - Unique IDs for requirements are now ensured  through standard validation.
  - Requirement Reference target IDs are are now ensured through standard validation.
  - Bi-directional requirement traces are now available, parent to child and child to parent.

Bug Fixes:
  - Circular imports no longer result in "infinite parsing" and a crash.

jondavid-black (commented):
Looks good to me.  The additional docstrings you put into model are very valuable.

Coffee2Bits (approved):

PR #529, opened 2023-02-01
Fixed bug 528

jondavid-black:
Simple change to check for key prior to accessing key in dict.
Also changed the way file names are created in the spec-csv command to align directly with input file rather than a potentially long name from within the file.


Coffee2Bits (approved):
Looks good. Is this a fix that you'd like to deploy today?

jondavid-black:
I would like to deploy rather quickly because right now I have to put in dummy requirements to work around the bug...which is really inconvenient.

jondavid-black:
One clarification...I removed the change to the output file name.  There really wasn't any reason to do anything here but explicitly fix the bug.

PR #535, opened 2023-02-02
Parallelized Nose2 Tests

### LINKED ISSUE #534: Run unit tests in parallel
As a developer, I want to speed up the pipeline execution so I get feedback faster.

Currently we run unit tests using tox which delegates to nose2.  We do not have it setup to run tests in parallel, but nose2 allows you do do that by using the following command `nose2 -c tox.ini --plugin=nose2.plugins.mp -N 10` to spawn 10 parallel threads for running tests.  A simple runtime comparison was done, by comparing the duration of the command from the last successful pipeline run, which took 511.019s, to a multi-threaded run using the command above in a 16 core GitPod environment, which took 121.636s co complete.  If this same speedup can be achieved in the pipelines, we'd cut ~6.5 minutes out of each unit test execution within the pipeline.  We would have to ensure there's nothing else going on that compromises the quality / stability of the tests and the coverage reporting.

AC:

- [ ] Configure unit tests to run in parallel.
- [ ] Ensure the tests are still run as expected and complete successfully.

Coffee2Bits:
Closes #534 

* Added configuration for parallelized tests

### DIFF
--- a/python/src/aac/plugins/first_party/lsp_server/providers/publish_diagnostics_provider.py
+++ b/python/src/aac/plugins/first_party/lsp_server/providers/publish_diagnostics_provider.py
@@ -1,10 +1,9 @@
 """Module for PublishDiagnostics Provider which handles publishing of diagnostics for AaC Language Server."""
-
-from typing import Optional
-
+import logging
 from pygls.lsp import Diagnostic, DiagnosticSeverity, PublishDiagnosticsParams
 from pygls.server import LanguageServer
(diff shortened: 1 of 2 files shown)


Coffee2Bits:
I'm a little concerned as I've seen intermittent failures with parallelizing the tests, but we can play it by ear as I've not been able to reproduce the failures.

jondavid-black (approved):
Looking forward to faster PRs!

PR #536, opened 2023-02-06
Primitive Number and Date Validation

### LINKED ISSUE #488: I want a primitive date validation
As an AaC user, I want a primitive validation for numbers, so that I can guarantee that fields with number values are valid decimal values.

Partially address #329 

## References:
[Validating an ISO 8601 datetime string in Python](https://stackoverflow.com/questions/41129921/validate-an-iso-8601-datetime-string-in-python)
[ISO 8601 Datetime Standard](https://en.wikipedia.org/wiki/ISO_8601)

## AC:
- [ ] There is a new primitive type validation for date
- [ ] The numbers primitive validation ensures that instances of "date" are valid ISO 8601 values.
- [ ] If a user puts a non-valid ISO 8601 value in a date field, their validation fails with an appropriate error message.

### LINKED ISSUE #490: I want a primitive number validation
As an AaC user, I want a primitive validation for numbers, so that I can guarantee that fields with number values are valid decimal values.

Partially address #329 

## References:

YAML Numbers Specification: https://symfony.com/doc/current/components/yaml/yaml_format.html#numbers

Apparently PyYAML has a bug with reading exponents: https://stackoverflow.com/a/30462009

## AC:
- [x] There is a new primitive type validation for numbers
- [x] The numbers primitive validation ensures that instances of "number" are numbers
  - [x] Integers are valid numbers
  - [ ] Octals are valid numbers
  - [x] Hexadecimals are valid numbers
  - [x] Floats are valid numbers
  - [x] Exponents are valid numbers
  - [ ] Infinity `.inf` is a valid number
- [x] If a user puts a non-numeric value in a number field, their validation fails with an appropriate error message.

Coffee2Bits:
Closes #490 
Closes #488 

* Added primitive number type validation
* Added primitive date type validation


jondavid-black (approved):
Look like reasonable validation tests to me.

PR #538, opened 2023-02-07
Added LSP handling for definitions with spaces in the name

### LINKED ISSUE #390: Enhance Goto Definition to find definitions with names that contain spaces
As an AaC user, I want Goto Definition to be able to find and take me to definitions that contain spaces in the names so that I can navigate my project files more flexibly.

## Notes:
Example definition with a space in the name:
https://github.com/jondavid-black/AaC/blob/3dfc11baf0ec8d7250608bd16caebdae837fc8af/python/src/aac/plugins/validators/defined_references/defined_references.yaml#L1-L2

Example reference to a definition with a space in the name:
https://github.com/jondavid-black/AaC/blob/3dfc11baf0ec8d7250608bd16caebdae837fc8af/python/src/aac/spec/spec.yaml#L117-L119

* It may be a helpful to treat newlines as delimiters for names that have spaces in the name.

## AC

- [x] Given a definition with spaces in the name (e.g. one of the validations) and a reference to that definition, when I activate a Goto Definition request from that reference, then I am taken to the definition.
  - [x] LSP logic for identifying the selected name is updated to account for spaces in the name
- [ ] There is a new references LSP test demonstrating this functionality.
- [ ] Refactor the `goto_definition_provider` module based on [this](https://github.com/jondavid-black/AaC/pull/377#discussion_r916334996) PR comment if it's still valid.

Coffee2Bits:
Closes #390 

* Added support for using LSP commands on definitions with spaces in their name


python/src/aac/plugins/first_party/lsp_server/providers/symbols.py
```diff
@@ -76,18 +76,25 @@ def get_symbol_range_at_position(content: str, line: int, column: int) -> Option
             symbol_start = 0
             # Reversed range from the adjusted column to the 0 element in the line
             for i in range(adjusted_column, -1, -1):
-                if line_with_symbol[i].isspace():
+                if line_with_symbol[i] == ':':
```
cvchaparro, line 79:
```suggestion
                if line_with_symbol[i] == ":":
```

python/src/aac/plugins/first_party/lsp_server/providers/symbols.py
```diff
@@ -76,18 +76,25 @@ def get_symbol_range_at_position(content: str, line: int, column: int) -> Option
...
                     break
 
                 symbol_start = i
 
             symbol_end = len(line_with_symbol)
             for i in range(adjusted_column, len(line_with_symbol)):
-                if line_with_symbol[i].isspace():
+                if line_with_symbol[i] == ':':
```
cvchaparro, line 86:
```suggestion
                if line_with_symbol[i] == ":":
```

python/tests/plugins/lsp_server/providers/test_goto_definition_provider.py
```diff
@@ -110,23 +110,22 @@ async def test_handles_goto_definition_request_when_definition_in_different_docu
         new_test_document_content = f"{TEST_DOCUMENT_CONTENT}{YAML_DOCUMENT_SEPARATOR}{TEST_SERVICE_THREE.to_yaml()}"
         await self.write_document(TEST_DOCUMENT_NAME, new_test_document_content)
 
-        target_location = self.get_definition_location_of_name(new_test_document_content, TEST_SCHEMA_C.name)
         res: GotoDefinitionResponse = await self.goto_definition(
-            TEST_DOCUMENT_NAME, line=target_location.range.start.line, character=target_location.range.start.character
+            TEST_DOCUMENT_NAME, line=61, character=17
```
cvchaparro, line 114: Hard-coding these values doesn't seem ideal since a change to the test document content (or more likely, a change to any of the definitions that make up the test document content) could change the specific `line` and `character` causing this test to fail.

cvchaparro (changes requested):
Minor feedback to start with. I am also debugging another issue where I'm seeing the hover text displayed twice in the hover box. I'll update you when I determine the cause.

*Update*: I re-generated the OpenAPI spec and re-built the extension and that resolved the issue.

python/tests/plugins/lsp_server/providers/test_goto_definition_provider.py
Coffee2Bits: I agree, but the logic I would have to implement to dynamically determine those lines would resemble the logic it's intended to test.

PR #543, opened 2023-02-09
Gen Plugin Adds AaC Version to Generated Plugins

### LINKED ISSUE #465: I want Gen-Plugin to generate the AaC version dependency based on the package's version
As an AaC plugin developer, I want Gen-Plugin to generate the AaC version dependency based on the package's version, so that my AaC dependency is properly versioned.

Current output in setup.py:
```
runtime_dependencies = [
    "aac"
]
```

Expected output in setup.py (Version 0.1.4 assumed):
```
runtime_dependencies = [
    "aac >= 0.1.4" 
]
```

## AC:
- [ ] Gen-plugin generates the aac runtime version with '>=' the current AaC version.

Coffee2Bits:
Closes #465

* Added support for generating plugins with the AaC package version.
* Cleaned up Gen Plugin bit
* Addressed some verbose nose2 test errors


jondavid-black (approved):
Lots of changes in gen-plugin command, but just cleanup stuff that looks good.  I'm good with the changes here.

PR #546, opened 2023-02-11
544 gen bom

jondavid-black:
This one ended up beign a little more than anticipated when I realized the best way to reduce complexity was to use validators to ensure material references were valid and there were no circular references present.

Changes:
- Modified the material-model plugin to add a gen-bom command.
- Added a validator to prevent circular references in material models.
- Added a validator to prevent bad references in material models.
- Added ability to reference requirement specs in deployments and assemblies.
- Added unit tests for all.
- Ensured no flake8 findings.


jondavid-black:
The failing test appears to be related to the state.json file.  If that file is present, running any aac command (even version) results in the `aac.lang.language_error.LanguageError: Duplicate target definitions found (2) with name 'Deployment'.` error.  If I delete the state.json file and run the command again, it works fine.  Then the state.json file reappears and aac commands are broken again.

I don't kiow what root cause is yet, but this is what I'm observing.

jondavid-black:
I did some local testing and found that the error is intermittent between indicating duplicate `Deployment` and `Assembly`.  I found that it was actually occurring in `language_context.py` line 646 when the `specification` plugin was being activated.  Perhaps there's some issue with using the `ext` schema across first-party plugins with the new context manager.  It still isn't clear, so I just removed the ability to reference requirements from a material model for now.

jondavid-black:
Looks like this may have just been a flakey test run for `tests.plugins.lsp_server.test_server.TestLspServer.test_adds_definitions_when_opening_file`.  I got the pipeline failure resolved by just rerunning it.

Coffee2Bits:
I'm pretty front-loaded this week, I'll have to circle back to this in a few days.

Coffee2Bits (approved):
Looks good!

PR #561, opened 2023-02-23
Validator KW Args and Core Spec relocation

### LINKED ISSUE #454: I want the Subcomponent Type validator overhauled to use keyword arguments for type and field
As an AaC user, I want the Subcomponent Type validator overhauled to use keyword arguments for type and field, so that I can specify type constraints for individual fields.

## References: 
[Subcomponent Validator](https://github.com/jondavid-black/AaC/blob/3dfc11baf0ec8d7250608bd16caebdae837fc8af/python/src/aac/plugins/validators/subcomponent_type/_subcomponent_type.py)

Current validator:
https://github.com/jondavid-black/AaC/blob/3dfc11baf0ec8d7250608bd16caebdae837fc8af/python/src/aac/spec/spec.yaml#L393-L394

Expected validator:
```yaml
  validation:
    - name: Field is type
      kw_args:
        - name: components
          type: model
```

Interface to update for KW arguments.
https://github.com/jondavid-black/AaC/blob/3dfc11baf0ec8d7250608bd16caebdae837fc8af/python/src/aac/validate/_validate.py#L150

Example plugin validator signature to update: https://github.com/jondavid-black/AaC/blob/3dfc11baf0ec8d7250608bd16caebdae837fc8af/python/src/aac/plugins/validators/root_keys/_validate_root_keys.py#L12-L17

## AC:
- [ ] Definition validations are updated to provide support for keyword arguments
  - [ ] Users can use list arguments or KW arguments
- [ ] The validator documentation is updated to include using keyword arguments
- [ ] The Subcomponent Type validator is updated to use keyword arguments
  - [ ] The key is the field to validate
  - [ ] The value defines what the type must be
- [ ] The Subcomponent Type validator is renamed to "Field Type"
- [ ] All first-party definition validator validation function signatures are updated with the keyword args.

### LINKED ISSUE #511: I want the Core Spec to be moved into the `aac.lang` package
I want the Core Spec to be moved into the `aac.lang` package.

## AC:

1. The core spec is moved into the `aac.lang` package.
2. References to the core spec are updated
3. References to the core spec in the docs are updated

Coffee2Bits:
Closes #454 
Closes #511 

* Moved the core spec module from `aac.spec.core` to `aac.lang.spec`
* Added support for keyword arguments for definition validations
* Updated Subcomponent Type validator to use keywords for uses to define different acceptable types for different fields


python/src/aac/lang/constants.py
```diff
@@ -17,6 +17,7 @@
 # Common Definition Field Names
 DEFINITION_FIELD_ADD = "add"
 DEFINITION_FIELD_ARGUMENTS = "arguments"
+DEFINITION_FIELD_KW_ARGUMENTS = "kw_arguments"
```
cvchaparro, line 20: I think it may be better to keep this following the Python "standard" where we say:

```suggestion
DEFINITION_FIELD_ARGUMENTS = "args"
DEFINITION_FIELD_KW_ARGUMENTS = "kwargs"
```

Alternatively, although less desirable IMO, would be to spell everything out completely if we're going to start off with separate fields for an "args" tuple - i.e. arguments - and an "args" dictionary - i.e. keyword_arguments.

python/src/aac/lang/definitions/lexeme.py
```diff
@@ -28,3 +28,7 @@ def __eq__(self, __o) -> bool:
             and self.location == __o.location
             and is_same_file(self.source, __o.source)
         )
+
+    def __hash__(self) -> int:
+        """Return the hash of this Definition."""
```
cvchaparro, line 33:
```suggestion
        """Return the hash of this Lexeme."""
```

python/src/aac/lang/definitions/source_location.py
```diff
@@ -19,6 +19,10 @@ class SourceLocation:
     position: int = attrib(validator=validators.instance_of(int))
     span: int = attrib(validator=validators.instance_of(int))
 
+    def __hash__(self) -> int:
+        """Return the hash of this Definition."""
```
cvchaparro, line 23:
```suggestion
        """Return the hash of this SourceLocation."""
```

python/src/aac/plugins/validators/_validator_findings.py
```diff
@@ -39,6 +40,14 @@ def _add_finding_with_severity(
...
         location = FindingLocation.from_lexeme(validation_name, lexeme)
         self.add_findings([ValidatorFinding(definition, severity, message, location)])
+        log_message = self._get_last_finding()
+
+        if severity == FindingSeverity.INFO:
+            logging.info(log_message)
+        elif severity == FindingSeverity.WARNING:
+            logging.warning(log_message)
+        elif severity == FindingSeverity.ERROR:
+            logging.error(log_message)
```
cvchaparro, line 50: I'm inclined to actually log outside of this module as that would allow more flexibility in how logging is done at the borders of the application (e.g. in the CLI) rather than in the internals.

python/src/aac/plugins/validators/_validator_findings.py
```diff
@@ -59,3 +68,10 @@ def get_info_findings(self) -> list[ValidatorFinding]:
     def _get_findings_with_severity(self, severity: FindingSeverity) -> list[ValidatorFinding]:
         """Return a list of findings of the specified severity."""
         return [finding for finding in self.findings if finding.severity == severity]
+
+    def _get_last_finding(self) -> ValidatorFinding:
+        return self._convert_finding_to_log_message(self.findings[-1])
+
+    def _convert_finding_to_log_message(self, finding: ValidatorFinding) -> str:
+        loc = finding.location
+        return f"Validation {finding.severity.name} failed for definition '{finding.definition.name}' from '{loc.validation_name}' in {loc.source.uri}:{finding.location.location.line}. {finding.message}"
```
cvchaparro, line 77: As mentioned above, I think this should be further from the internals. If I want my finding/log messages to be formatted differently or if I want to include different/additional information it would be challenging, like this.

python/src/aac/plugins/validators/reference_type/_reference_type.py
```diff
@@ -0,0 +1,70 @@
...
+
+PLUGIN_NAME = "Reference is type"
+
+
+def validate_reference_types(
+    definition_under_test: Definition,
+    target_schema_definition: Definition,
+    language_context: LanguageContext,
+    *validation_args,
+    **validation_kw_args,
```
cvchaparro, line 15: Python usually uses `**kwargs` to represent keyword arguments, I think it would be good to keep in line with that style.

```suggestion
    **validation_kwargs,
```

cvchaparro (commented):
I tested the changes and everything seems to work as expected.

lizzcondrey:
This work is stale and was reworked within the 0.4.0 baseline of #714 . Closing this out.

PR #542, opened 2023-04-11
I want a primitive reference validation

### LINKED ISSUE #550: I want a primitive string validation
As an AaC user, I want a primitive validation for string, so that I can guarantee that string values are actually strings.

There is a question of how do we validate strings; the goal for this validator is to ensure that fields of type 'string' are represented as strings in the python environment. This will help us to avoid issues where plugin developers are expecting to be able to perform python string operations on a field that is designated as a string with the confidence that the value won't be a non-string object.

## References:

## AC:
- [ ] There is a new primitive type validation for strings
- [ ] The string primitive validation ensures that string are instances of python string objects.
- [ ] If a user puts a non-valid value in a string field, their validation fails with an appropriate error message.
  - [ ] The message must include a note about surrounding the offending value with quotations to turn it into a string value.

cvchaparro:
Note: Also closes #550.

As an AaC user, I want a primitive validation for references, so that I can guarantee that references are valid.

## References:

Validation to remove: https://github.com/jondavid-black/AaC/blob/e743b99051aa96086a8ecc6fc5f7a1ff531b1d7b/python/src/aac/spec/spec.yaml#L117-L119

Field to update: https://github.com/jondavid-black/AaC/blob/e743b99051aa96086a8ecc6fc5f7a1ff531b1d7b/python/src/aac/spec/spec.yaml#L93-L94

## AC:
- [x] There is a new primitive type validation for reference
- [x] The references primitive validation ensures that references are instances of definitions.
- [x] If a user puts a non-valid value in a reference field, their validation fails with an appropriate error message.
- [x] The extension definition has the `Type references exist` validation removed
- [x] The extension definition's field 'type' is changed from a string to a reference primitive


Coffee2Bits (approved):
Fantastic!

PR #581, opened 2023-04-12
Fix .aac files being opened in the visual editor

### LINKED ISSUE #540: AaC Extension opens .aac files in the Visual Editor regardless of mode
AaC Extension opens .aac files in the Visual Editor regardless of mode. If a user has not engaged the AaC REST Api plugin, or selected the special AaC editor in the activity bar, the plugin will still open .aac files in the visual editor window.

## AC:
- [ ] The visual editor windows are only opened when a .aac file is opened via the AaC Visual Editor activity bar/view.
- [ ] .aac files should be opened as yaml files if the visual editor isn't active (No connection to the RESTful app.)

Coffee2Bits:
Closes #540

Turns out the problem was mostly straightforward, though I couldn't get the visual editor to open unless I had a non-empty string in the file selector value. 

Visual of the fix:
image

### DIFF
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -197,7 +197,7 @@
                 "displayName": "Edit AaC Definitions",
                 "selector": [
                     {
-                        "filenamePattern": "*.aac"
+                        "filenamePattern": "."
                     }
                 ],
                 "priority": "default"


cvchaparro (approved):
Seems to work; nice!

PR #500, opened 2023-04-26
I want a `Plugin` root key and schema

cvchaparro:
As a plugin developer, I want a dedicated root key and schema for defining plugins, so that I have a clearer and more explicit definition of what my plugin does.

## AC:
- [x] In the core spec there is a new schema defined `Plugin`
- [x] In the core spec, root keys defines a new entry `plugin` with the new `Plugin` type
- [x] ~There is a new `PluginContributions` enum which provides a list of enums for each of the plugin contribution types.~
- [x] There is a new schema `PluginCommand` that has the following fields:
  - [x] Group - A string indicating the command's group name
  - [x] Input
  - [x] Output
- [x] The new schema `Plugin` has the following fields
  - [x] name - The name for the plugin
  - [x] commands - An optional list of `Behavior` entries
  - [x] additional_definition_resources - An optional list of files that contain the plugin-provided definitions
  - [x] primitive_validations - An optional list of primitive-type validators
  - [x] definition_validations - An optional list of definition-type validators
- [x] The Gen-Plugin plugin is updated to use the new schema
- [x] Existing plugins have their definitions updated


python/tests/helpers/io2/TemporaryTestFile.py
```diff
@@ -0,0 +1,83 @@
...
+        Return a TemporaryTestFile context-manager object.
+
+        Arguments:
+            content (str): A string to use as the contents of the file.
+            name (Optional[str]): (default: None)
+            clean_up (Optional[str]): Whether to clean up the test file/directory after exiting the context manager.
+              Recognized Options:
+                'all': Remove `dir` along with all files and directories under `dir`. (default)
+                'files': Remove all files under `dir` only.
+                'directories': Remove all directories under `dir` only.
```
cvchaparro, line 33: TODO

python/tests/helpers/io2/TemporaryTestFile.py
cvchaparro, line 1: I'd like to update the usages of `temporary_test_file` to use `TemporaryTestFile` as we can specify names for the test files more easily, this way. I didn't think it was worth making this change in the PR, though as it's big enough already.

Coffee2Bits (changes requested):
Looks good. Instead of the helper IO and IO2 modules, can we rename IO -> IO_DEP and IO2-> IO?

python/src/aac/plugins/contributions/contribution_types/definition_validation_contribution.py
```diff
@@ -8,17 +12,20 @@
...
-        validation_function: The validation callback function
+        name (str): A string with the name of the command argument.
+        definition (str): The AaC definition of the Validator Plugin.
+        validation_function (Callable): The validation callback function.
     """
 
     name: str = attrib(validator=validators.instance_of(str))
     definition: Definition = attrib(validator=validators.instance_of(Definition))
-    validation_function: callable = attrib(validator=validators.is_callable())
+    validation_function: Callable = attrib()
```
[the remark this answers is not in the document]
cvchaparro: The `is_callable` validator was causing issues when implementing our fix for the "only require validator implementations for active plugins" because the check in our validator required seeing if the `validation_function is None` but the `is_callable` validator doesn't allow `None` for a function.

python/src/aac/plugins/contributions/contribution_types/primitive_validation_contribution.py
```diff
@@ -1,22 +1,29 @@
...
-        validation_function: The validation callback function
+        name (str): The name of the enum validator
+        primitive_type (str): The enum type
+        validation_function (Callable): The validation callback function
     """
 
     name: str = attrib(validator=validators.instance_of(str))
     primitive_type: str = attrib(validator=validators.instance_of(str))
-    validation_function: callable = attrib(validator=validators.is_callable())
+    validation_function: Callable = attrib()
```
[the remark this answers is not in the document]
cvchaparro: See other response.

python/src/aac/plugins/contributions/contribution_types/definition_validation_contribution.py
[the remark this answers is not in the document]
Coffee2Bits: Ah, this can be accomplished by wrapping the is_callable call with the optional function like so:
```    validation_function: Optional[Callable] = attrib(validator=validators.optional(validators.is_callable()))``` For some reason I'm unable to create a suggestion for this line.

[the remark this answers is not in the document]
Comment (cvchaparro) [reply python/src/aac/plugins/contributions/contribution_types/definition_validation_contribution.py:28]: Okay, I'll use that - thanks!

File: python/tests/helpers/io2/TemporaryTestFile.py
Comment (cvchaparro) [reply python/tests/helpers/io2/TemporaryTestFile.py:1]: Per my commit message for commit b5abbdb, I went ahead and updated this (since I had to touch the files for package updates anyways).

File: python/src/aac/plugins/contributions/contribution_types/definition_validation_contribution.py [hunk 4]
```diff
@@ -8,17 +12,20 @@
...
+    A class that contains all the relevant information to manage and execute definition validation
+    contributions with the validation process.
 
     Attributes:
-        name: A string with the name of the command argument
-        definition: The AaC definition of the Validator Plugin
-        validation_function: The validation callback function
+        name (str): A string with the name of the command argument.
+        definition (str): The AaC definition of the Validator Plugin.
+        validation_function (Callable): The validation callback function.
```
cvchaparro, line 23:
```suggestion
        validation_function (Optional[Callable]): The validation callback function.
```

python/src/aac/plugins/contributions/contribution_types/primitive_validation_contribution.py
```diff
@@ -1,22 +1,29 @@
...
+    A class that contains all the relevant information to manage and execute type validator
+    contributions against type-value enums in the validation process.
 
     Attributes:
-        name: The name of the enum validator
-        primitive_type: The enum type
-        validation_function: The validation callback function
+        name (str): The name of the enum validator
+        primitive_type (str): The enum type
+        validation_function (Callable): The validation callback function
```
cvchaparro, line 21:
```suggestion
        validation_function (Optional[Callable]): The validation callback function
```

python/tests/helpers/io/TemporaryTestFile.py
```diff
@@ -0,0 +1,97 @@
...
+
+        Arguments:
+            content (str): A string to use as the contents of the file.
+            name (str): An optional name to give to the file. (default: "")
+            clean_up (str): Whether to clean up the test file/directory after exiting the context manager.
+              Recognized Options:
+                'all': Remove `dir` along with all files and directories under `dir`. (default)
+                'files': Remove all files under `dir` only.
+                'directories': Remove all directories under `dir` only.
+                'none': Do not remove any files/directories under `dir`.
```
Coffee2Bits, line 34: We should use an enum here.

Coffee2Bits (changes requested):
Looks good. Only change would be to use enums to select the Temp file mode instead of strings.

Coffee2Bits (approved):
Great work!

PR #595, opened 2023-05-01
Adding PR Template

lizzcondrey:
Added PR template to .github directory for central repository resources so that it is accessible for leverage within the AaC prototype project repository.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -30,7 +30,6 @@ Include pertinent changes that are improvements to existing features.
 - None
 
 # Checklist:
-- [ ] Code is commented (at minimum new code, but if legacy code is missing comments, try to update.)
 - [ ] Updated documentation accordingly.
 - [ ] My changes generate no new warnings.
 - [ ] Have updated new and existing unit tests accordingly.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,9 +2,9 @@
 Please include a summary of the changes and the related issue. List any dependencies that are required for this change.
 
 # Linked Items:
-Please include the associated items of this Pull Request using the following syntax to link them for closure on completion of this Pull Request.
-`Closes #<Issue Number>`
-`Fixes #<Issue Number>`
+Please include the items resolved by and/or associated with this Pull Request using the following syntax to link them for closure upon completion.
+`Closes #<Issue Number>` or
+`Fixes #<Issue Number>` or
 `Resolves #<Issue Number>`
 
 ## Type of change
@@ -12,7 +12,6 @@ Please include the associated items of this Pull Request using the following syn
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [ ] Feature improvement
-- [ ] Requires a documentation update
 
 ### Feature Additions
 Include pertinent changes as a result of adding the new feature.
@@ -31,11 +30,8 @@ Include pertinent changes that are improvements to existing features.
 - None
 
 # Checklist:
-- [ ] Code follows the style guidelines of this project.
 - [ ] Code is commented (at minimum new code, but if legacy code is missing comments, try to update.)
 - [ ] Updated documentation accordingly.
 - [ ] My changes generate no new warnings.
 - [ ] Have updated new and existing unit tests accordingly.
-- [ ] Unit tests pass locally.
-- [ ] Pipeline completes and passes all stages.
 - [ ] Linked associated items to be closed.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -23,7 +23,7 @@ Include pertinent changes of fixing the bug.
 - None
 
 ### Breaking Changes
-Include pertinent changes that qualify as breaking changes. Be sure to include information of the impact to end users and additional information for addressing the break if more than a version transition is necessary for mitigating. 
+Include pertinent changes that qualify as breaking changes. Be sure to include information about the impact to end users and additional information for addressing the break if more than a version transition is necessary for mitigating. 
 - None
 
 ### Feature Improvement Changes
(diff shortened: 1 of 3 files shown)


Coffee2Bits:
Looks like the wheel version change is affecting you here. 

The fix is to change this line https://github.com/jondavid-black/AaC/blob/9232c42406ae212fbe4e42905936fcf591c79c11/python/setup.py#L37 to "wheel ~= 0.40.0"

.github/pull_request_template.md
```diff
@@ -0,0 +1,41 @@
...
+
+# Checklist:
+- [ ] Code follows the style guidelines of this project.
+- [ ] Code is commented (at minimum new code, but if legacy code is missing comments, try to update.)
+- [ ] Updated documentation accordingly.
+- [ ] My changes generate no new warnings.
+- [ ] Have updated new and existing unit tests accordingly.
+- [ ] Unit tests pass locally.
+- [ ] Pipeline completes and passes all stages.
+- [ ] Linked associated items to be closed.
```
Coffee2Bits, line 41: I think we can slim this down since we have automation to check for styling, unit tests passing locally (we already cover all major operating systems in the pipeline), pipeline passing will automatically block the PR, and linked items should be automatically closed out on this PR is closed.

```suggestion
- [ ] Code is commented (at minimum new code, but if legacy code is missing comments, try to update.)
- [ ] Updated documentation accordingly.
- [ ] Have updated new and existing unit tests accordingly.
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,41 @@
+# Description
+Please include a summary of the changes and the related issue. List any dependencies that are required for this change.
+
+# Linked Items:
+Please include the associated items of this Pull Request using the following syntax to link them for closure on completion of this Pull Request.
```
cvchaparro, line 5:
```suggestion
Please include the items resolved by and/or associated with this Pull Request using the following syntax to link them for closure upon its completion.
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,41 @@
...
+`Closes #<Issue Number>`
+`Fixes #<Issue Number>`
+`Resolves #<Issue Number>`
```
cvchaparro, line 8:
```suggestion
`Closes #<Issue Number>` or
`Fixes #<Issue Number>` or
`Resolves #<Issue Number>`
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Type of change
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- [ ] Feature improvement
+- [ ] Requires a documentation update
```
cvchaparro, line 15: We may want to start requiring documentation updates for most changes, except possibly minor bug fixes, etc so this may need to be an automatic assumption.

.github/pull_request_template.md
```diff
@@ -0,0 +1,41 @@
...
+Include pertinent changes that qualify as breaking changes. Be sure to include information about the impact to end users and additional information for addressing the break if more than a version transition is necessary for mitigating. 
+- None
+
+### Feature Improvement Changes
+Include pertinent changes that are improvements to existing features.
+- None
+
+# Checklist:
+- [ ] Code follows the style guidelines of this project.
+- [ ] Code is commented (at minimum new code, but if legacy code is missing comments, try to update.)
```
cvchaparro, line 35: I'd favor improving the readability of the code over adding comments as a way to try to describe what's going on so I would suggest we either remove this or update to say that adding comments is needed when it can't be made self-documenting.

cvchaparro (changes requested):

.github/pull_request_template.md
lizzcondrey: Requested removals make sense. The list was generated without having this project pipeline up as reference, so tried to cover the things that I've seen fall off on teams. 

Would argue to keep the linked items piece as a follow up for the section earlier in the template to help ensure it gets done since there aren't backup automation checks in for it yet. While the linked items will automatically be closed with completion of the PR, there isn't an assist for making sure items were linked to be closed.

.github/pull_request_template.md
lizzcondrey: Can remove the `Requires documentation update` from the type of change section and leave it in the final check list about `Updated documentation accordingly` as a reminder.

Coffee2Bits: I think it's fine to leave as a reminder to check for documentation updates.

.github/pull_request_template.md
Coffee2Bits: I'm split on this since just relying on self-documenting code can become very technical and be difficult for newer or less experienced developers whereas a comment can easily provide high-level context to the reader. Since this is an OS project, I'm inclined for us to spend some cycles supporting and maintaining high-level clarifying comments where we can.

lizzcondrey: Middle ground of updating comments as you touch things and adding `TODO` references with issue numbers for code readability refinement and self-documenting improvements? 

Would see that as being two separate check list items.

cvchaparro: > Can agree to that preference. But that type of change, to me, would likely fall outside of the original scope of the task. Whereas updating comments is easier to cover within scope from what I have usually seen in previous project work.

I don't really see what you mean by "fall outside the original scope of the task". The suggestion is to clarify the comment requirement to say something like "add comments as needed".

> I'm split on this since just relying on self-documenting code can become very technical and be difficult for newer or less experienced developers whereas a comment can easily provide high-level context to the reader. Since this is an OS project, I'm inclined for us to spend some cycles supporting and maintaining high-level clarifying comments where we can.

To clarify, I'm not saying don't add comments at all just opt for code clarity over comments. I've seen it happen where comments are used in place of writing descriptive code and I am strongly opposed to that. That said, could you give an example of what would be more complex when reading self-documenting code (which is known to be up-to-date) over descriptive comments (which can often be out-of-date)? In general, I'd rather spend cycles updating code to be more descriptive (i.e. a descriptive function name with a docstring is 1000x better than a descriptive comment, IMO).

> Middle ground of updating comments as you touch things and adding `TODO` references with issue numbers for flagging areas eligible for code readability refinement and self-documenting improvements?
> 
> Would see that as being two separate check list items.

TODO references will break the linter and hence the pipeline (unless we turn that warning off).

lizzcondrey: > I don't really see what you mean by "fall outside the original scope of the task". The suggestion is to clarify the comment requirement to say something like "add comments as needed" but not require them to be added to all new/updated code which will get superfluous very quickly.

The scope comment was with regards to re-working the code to be more readable and self-documenting, not about modifying the checklist item. My referencing in my response was unclear. 

> TODO references will break the linter and hence the pipeline (unless we turn that warning off).

Good to know, I have not dug through the pylint and flake8 configurations to see what all is currently set. 

Definitely a good discussion piece for me to learn how you all handle referencing issues within the code.

Coffee2Bits: Let's remove the code commenting bit for now. I expect reviewers to provide feedback on whether or not the PR content is readable and documented.
```suggestion
```

Coffee2Bits (approved):

cvchaparro (approved):
Nice job!

PR #597, opened 2023-05-02
Root Key Rename Duplicates Colon

### LINKED ISSUE #539: Root Key Rename Duplicates colons
There is an issue between the LSP client and our LSP python server when renaming YAML keys adjacent to colons; the colon becomes duplicated during rename operations.

## AC:
- [ ] Renaming YAML keys adjacent to colons does not produce duplicate colons.

Coffee2Bits:
Closes #539

* Fixed the issue of duplicate colons by stripping colons in rename text provided by the LSP client
* Added uuid logic to definitions to address some duplication issues in the language context

# Evidence
Screenshots of the model root key being renamed to 'my_model' demonstrating that there is no extra colon appended to the end (e.g. `my_model::`).
Screenshot 2023-05-01 at 11 53 56 PM
Screenshot 2023-05-01 at 11 54 20 PM

### DIFF
--- a/python/src/aac/__init__.py
+++ b/python/src/aac/__init__.py
@@ -12,7 +12,7 @@ if sys.version_info < (3, 9):
 import logging
 import os
 
-__version__ = "0.1.10"
+__version__ = "0.1.11"
 __log_file_name__ = os.path.join(os.path.dirname(__file__), "aac.log")
 
 logging.basicConfig(
--- a/python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
+++ b/python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
@@ -173,5 +173,3 @@ class RenameProvider(LspProvider):
 def _filter_editable_definitions(definitions_to_filter: List[Definition]) -> List[Definition]:
     filtered_definitions = {definition for definition in definitions_to_filter if definition.source.is_user_editable}
     return list(filtered_definitions)
-
-
--- a/python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
+++ b/python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
@@ -66,11 +66,12 @@ class RenameProvider(LspProvider):
 
     def _get_rename_edits(self, symbol: str, new_name: str) -> dict[str, list[TextEdit]]:
         """Returns the list of text edits based on the selected symbol's type."""
-        language_context = self.language_server.language_context
 
         # Strip any ':' that accompany yaml keys from the incoming replacement text
         sanitized_new_name = new_name.strip(":")
-        text_to_replace = remove_list_type_indicator(symbol).strip(":")
+        text_to_replace = remove_list_type_indicator(symbol)
+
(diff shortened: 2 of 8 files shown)


python/src/aac/plugins/first_party/lsp_server/providers/symbols.py
```diff
@@ -81,7 +81,7 @@ def get_symbol_range_at_position(content: str, line: int, column: int) -> Option
 
                 symbol_start = i
 
-            symbol_end = len(line_with_symbol)
+            symbol_end = len(line_with_symbol) - 1
```
Coffee2Bits, line 84: This was an unrelated off-by-one error that I've seen cropping up for a bit.

python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
```diff
@@ -65,38 +66,37 @@ def get_rename_edits(
...
-        if not symbol:
-            return None
-
         language_context = self.language_server.language_context
 
-        name = remove_list_type_indicator(symbol).strip(":")
-        symbol_types = get_possible_symbol_types(name, language_context)
+        # Strip any ':' that accompany yaml keys from the incoming replacement text
+        sanitized_new_name = new_name.strip(":")
+        text_to_replace = remove_list_type_indicator(symbol).strip(":")
```
cvchaparro, line 73: Slight improvement but we should probably follow the `remove_list_type_indicator` pattern and put the `strip(":")` into a simple helper function, as well.

cvchaparro (changes requested):
Can you please add a unit test that covers the rename with a colon?

python/src/aac/plugins/first_party/lsp_server/providers/rename_provider.py
Coffee2Bits: I adjusted the logic so that we're only stripping `new_name`. I don't foresee the need to do a lot of stripping like this -- it's necessary because of how VSCode's rename function incorporates the colon in the request. So, TLDR: I don't see it as necessary to provide a function to wrap `.strip(':')` here.

cvchaparro (changes requested):

cvchaparro (approved):

python/src/aac/__init__.py
```diff
@@ -12,7 +12,7 @@
 import logging
 import os
 
-__version__ = "0.1.10"
+__version__ = "0.1.11"
```
lizzcondrey, line 15: Maybe that's worth an edit for the PR template, reminder about bumping appropriate package versions for changes?

Coffee2Bits: Hmmm that's a good question. It could be helpful to have a reminder, but every PR won't have a version bump since we have multiple PRs per release at the moment.

lizzcondrey: Per conversation in scrum today we reached the agreement of putting a reminding comment of considering if a version bump is necessary within the PR template. And since it's a one line text update it would be okay to put in with this PR.

lizzcondrey (approved):

PR #612, opened 2023-06-05
Make the import keyword a file wide declaration

### LINKED ISSUE #582: Fix DocumentLink provider test on MacOS
One of the `DocumentLink` provider tests is failing on MacOS with the following error:

``` shellsession
cd python; flake8 . && nose2 -c tox.ini
0
ERROR:root:Failed to find the target definition 'root' in the context.
ERROR:root:Failed to find the target definition 'root' in the context.
........................................................................................................................................................................................................................................................Successfully added /private/var/folders/11/zhn1zq0j4ts6bqw_y2tp39xw0000gq/T/tmpfoctifmw/tmp7trtrffy to the active context
................................F.................................................

======================================================================
FAIL: test_document_link_provider (tests.plugins.lsp_server.providers.test_document_link_provider.TestDocumentLinksProvider)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.10/3.10.10_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/async_case.py", line 64, in _callTestMethod
    self._callMaybeAsync(method)
  File "/opt/homebrew/Cellar/python@3.10/3.10.10_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/async_case.py", line 87, in _callMaybeAsync
    return self._asyncioTestLoop.run_until_complete(fut)
  File "/opt/homebrew/Cellar/python@3.10/3.10.10_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/opt/homebrew/Cellar/python@3.10/3.10.10_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/async_case.py", line 101, in _asyncioLoopRunner
    ret = await awaitable
  File "/Users/user/repo-location/aac/python/tests/plugins/lsp_server/providers/test_document_link_provider.py", line 43, in test_document_link_provider
    res: DocumentLinkResponse = await self.request(text_document=new_file_path)
  File "/Users/user/repo-location/aac/python/tests/plugins/lsp_server/providers/test_document_link_provider.py", line 27, in request
    return await self.build_request(
  File "/Users/user/repo-location/aac/python/tests/plugins/lsp_server/base_lsp_test_case.py", line 181, in build_request
    self.assertIsNotNone(
AssertionError: unexpectedly None : Could not execute textDocument/documentLink in virtual document because there is no document named /var/folders/11/zhn1zq0j4ts6bqw_y2tp39xw0000gq/T/tmp3x084zbn/new.aac.

----------------------------------------------------------------------
Ran 330 tests in 57.974s

FAILED (failures=1)

# Code coverage data elided for brevity

99 files skipped due to complete coverage.
```

### LINKED ISSUE #600: I want the `import` keyword  moved from definitions to a file-wide declaration
As an AaC user, I want the `import` keyword  moved from definitions to a file-wide declaration, because import clauses associated with individual definitions satisfy import dependencies for all other definitions in the file making it a potentially confusing approach.

### Discussed in https://github.com/jondavid-black/AaC/discussions/569

AC:
- [x] Documentation pertaining to imports is updated accordingly
  - [x] Documentation is updated to make the file-wide scope explicit to the user
- [x] Existing example models and plugin definitions are updated accordingly
- [x] The parser is updated to store import information in the internal file class 'AaCFile'
  - [x] Import-specfic logic for definition parsing is removed

cvchaparro:
# Description
Change the way `import`s are treated such that we move it to a top-level definition as opposed to treating it as a special root that gets added to every definition and is treated specially during parsing.

# Linked Items:
Please include the items resolved by and/or associated with this Pull Request using the following syntax to link them for closure upon completion.
Closes #600
Resolves #582

## Type of change
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Feature improvement

### Feature Additions
- None

### Bug Fixes
- I changed the `tests` package initialization to ensure `TemporaryDirectory`s are always resolving symlinks on creation (as opposed to just during use as a context manager).

### Breaking Changes
- The `import` keyword is now considered it's own definition (against which validations, etc can be written) as opposed to a treating it specially during parsing.

- None

# Checklist:
- [x] Updated documentation accordingly.
- [ ] Have you bumped the version if there will be a release?


python/src/aac/lang/definitions/definition.py
```diff
@@ -127,47 +130,46 @@ def get_fields(self) -> Optional[list[dict]]:
         fields = self.get_top_level_fields()
         return fields.get(DEFINITION_FIELD_FIELDS)
 
-    def get_imports(self) -> list[str]:
-        """Return a list of imported aac files, empty if there are no imports."""
-        return self.imports
+    def get_imports(self) -> Optional[list[str]]:
+        """Return a list of imported files, or None if the definition is not import."""
+        imports = self.get_top_level_fields()
+        return imports.get(ROOT_KEY_IMPORT)
```
Coffee2Bits, line 136: * Note to self to review

python/src/aac/lang/definitions/definition.py
```diff
@@ -89,7 +90,9 @@ def get_top_level_fields(self) -> dict[str, dict]:
         """
         fields = self.structure.get(self.get_root_key())
 
-        if not fields:
+        if self.is_import():
+            fields = {ROOT_KEY_IMPORT: fields}
```
Coffee2Bits, line 94: I'm inclined to move this into the `get_imports` function for consistency. Was there a blocker to that pattern?

Coffee2Bits (changes requested):
Looks good! Just one question/comment.

python/src/aac/lang/definitions/definition.py
cvchaparro: I didn't try that so it may work but the issue I ran into was that the `import` definition doesn't have fields - it was returning the list of imports - which broke some places we called `get_top_level_fields` and it tried accessing a `get` method when there isn't one.

An alternative approach I can try is to put some extra stuff in the import definition structure attribute but it seemed a bit cleaner to do it this way, I thought.

Coffee2Bits (approved):

PR #619, opened 2023-06-15
Update Core Spec to adhere to the Style Guide

### LINKED ISSUE #602: I want the Core Spec updated to the AaC Style Guide
As an AaC user, I want the Core Spec updated to the style guide standards so that I have a consistent understanding of the target style between documentation and the standard library/core spec.

Blocked by #601 

The core specification (a.k.a standard library for AaC), can be found [here](https://github.com/jondavid-black/AaC/blob/main/python/src/aac/spec/spec.yaml).

## AC:
- [ ] The Core Spec is updated to the style guide standard
- [ ] Python references to core spec fields/components are updated accordingly
  - [ ] AaC python functions/user-features are updated
  - [ ] Unit tests are updated
- [ ] AaC files using core spec definitions are updated accordingly

Coffee2Bits:
# Description
Update the Core Spec to adhere to the style guide.

# Linked Items:
Please include the items resolved by and/or associated with this Pull Request using the following syntax to link them for closure upon completion.
Closes #602

## Type of change
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Feature improvement

### Feature Additions
- None

### Bug Fixes
- None

### Breaking Changes
- Updated the Core Spec to adhere to the style guide
  - The `schema` definition is now named `Schema`
  - The `model` definition is now named `Model`
  - The `root` definition is now named `Root`
  - The `enum` definition is now named `Enum`
  - The `extension` definition is now named `Extension`
  - The enum `BehaviorType`'s value are now UPPER_CASE
    - `pub-sub` is now `PUB_SUB`
    - `request-response` is now `REQUEST_RESPONSE`
    - `startup` is now `STARTUP`
    - `timer` is now `TIMER`
    - `logging` is now `LOGGING`
- Removed several old definitions from the Core Spec
  - `Map` was moved to Gen-Protobuf
  - `KeyValuePair` was moved to Gen-Protobuf
  - Removed `DefinitionReference` as its been replaced by the primitive type `reference`
- Updated First Party Plugins to adhere to the style guide

- None

# Checklist:
- [x] Updated documentation accordingly.
- [x] Have you bumped the version if there will be a release?


python/tests/plugins/lsp_server/providers/test_semantic_tokens_provider.py
```diff
@@ -61,7 +61,7 @@ async def test_get_semantic_tokens_for_enum(self):
 
         props = {
             enum.name: {
-                enum.get_root_key(): (SemanticTokenTypes.Keyword, 0b11),
+                enum.get_root_key(): (SemanticTokenTypes.Keyword, 0b10),
```
Coffee2Bits, line 64: Updated this to reflect the change in behavior in the token provider. I believe the 3 -> 2 change is because a number of these names no longer match the root key, which means it matches with one fewer of the bitmask checks.

cvchaparro: I think we still want this to be `0b11` as the bitmask indicates that a token is [readonly](https://github.com/jondavid-black/AaC/blob/e743b99/python/src/aac/plugins/first_party/lsp_server/providers/semantic_tokens_provider.py#L51-L58) and it is [defined in the default library](https://github.com/jondavid-black/AaC/blob/e743b99/python/src/aac/plugins/first_party/lsp_server/providers/semantic_tokens_provider.py#L61-L65) - i.e. defined in the core spec. As such, I think these root keys should always have both modifiers - i.e. `0b11`. I think we need a slight modification of the logic in those two functions (linked above).

cvchaparro (changes requested):
Good job, so far. I have a few comments.

python/tests/plugins/lsp_server/providers/test_semantic_tokens_provider.py
Coffee2Bits: Hmm. Why are we considering root keys as readonly?

cvchaparro (approved):
Nice!

PR #667, opened 2023-09-07
Create the Usecase participant field validator

### LINKED ISSUE #622: I want an AaC reset command
As an AaC user, I want an AaC reset command, so that I can get back to a clean state regardless of the error state I've gotten into.

Relates to #617 and #618 

## AC:
- [ ] There is an AaC plugin command 'reset'
- [ ] When 'reset' is run the `state.json` file is deleted
- [ ] This command pre-empts the language context or state management so that its execution can't be prevented

### LINKED ISSUE #662: 📋 [TASK] - Create Use Case Participant Field Validator
### Parent Feature

Expand Validators

### Task Description

As an AaC Developer, I want a use case participant field validator, so that I know I have valid references to use case participants within the use case definition. 

### Acceptance Criteria

- [ ] A new validator is created for use case participant field
- [ ] A unit test for the validator is created
    - [ ] Including a pass model and fail model for testing validator thoroughly
- [ ] The validator is added to the pipeline core validators if necessary
    - [ ] Script work for properly handling assert failures within pipeline to not cause pipeline failure but handle that model failed validation correctly
- [ ] The validator documentation is updated to cover the new validator
- [ ] Validator should check that participants are included as part of the use case participant field

### Unresolved questions

_No response_

cvchaparro:
# Description

Add a validator that checks the `usecase`s `step` definitions to ensure that the
`source` and `target` fields refer to valid `participant`s of the usecase.

# Linked Items:

Closes #662 

## Type of change
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Feature improvement

### Feature Additions
- New validator that checks `source` and `target` against the list of
  `participants` in the `usecase`.

### Bug Fixes
- None

### Breaking Changes
- None

- None

# Checklist:
- [ ] Updated documentation accordingly.
- [x] Have you bumped the version if there will be a release?


.dir-locals.el
```diff
@@ -17,6 +17,7 @@
                  (python-tab-width . 4)
```
lizzcondrey, line 17: First time I've seen us making changes to these files....or that I've seen as part of a project period (that I can recall, I don't remember having any `*.el` files in previous projects). They seem to be config-esque type files. Is that assumption correct? What are they for?

python/src/aac/plugins/validators/__init__.py
```diff
@@ -7,14 +7,16 @@
...
 
 __all__ = (
-    ValidatorResult.__name__,
-    ValidatorFinding.__name__,
-    ValidatorFindings.__name__,
-    FindingLocation.__name__,
-    FindingSeverity.__name__,
-    get_validation_definition_from_plugin_yaml.__name__,
-    get_validation_definition_from_plugin_definitions.__name__,
+    "ValidatorResult",
```
lizzcondrey, line 14: Why is it better to use the string literal rather than the `*.__name__` reference? With the `*.__name__` piece, if it gets changed in the code, we don't have to update here. But would if we used the string literals? If I am thinking about that correctly, anyway.

.dir-locals.el
cvchaparro: These have been in for a while. They're basically the Emacs equivalent to the Gitpod configs.

python/src/aac/plugins/validators/__init__.py
cvchaparro: If you're using the Pyright LSP server (I think this is what VSCode uses in the Python extension) it handles renames appropriately and this is the preferred Pythonic way to do it from what I've seen in other projects, docs, warnings on the `__all__` symbol from my editor, etc.

Given that it's handled appropriately (presumably by the main Python LSP servers, at least), I'd recommend we move to this style.

Coffee2Bits: The official MS Python plugin supports Pylance or Jedi for the LSP implementation.
Screenshot 2023-09-07 at 8 21 25 PM

cvchaparro: I guess that's what happens when I make assumptions about VSCode. 🙃 That said, while I generally don't like Microsoft stuff - pyright seems to work pretty well.

lizzcondrey: Haha, love that my sanity check spurred a small tooling discussion. 

So, for Cameron's initial comment back, agree the string literal is the general Pythonic way I have seen and previously used. But since this PR was swapping over from the `*.__name__` I took that to mean that those components had possibly previously been often to change. Which I suppose is a good reason to swap to the literal if they don't change often in truth. 

`__all__` can be finicky with linters from what I have seen in the past. It is a useful exposure management component for your packages but can be funky to judge proper inclusion sometimes. Definitely still an area I am a bit shaky in my understanding. 

To Alex's comment, agree that standardizing on implementation and tooling would be useful so that it is easier to provide support to each other when we encounter issues. Also, would help for clarifying what support we can provide to users for issues they have.

lizzcondrey:
@cvchaparro is this PR ready for full review? You had sent a status yesterday about wanting to do some reworks, but I haven't seen a commit around that. So, I just hit pieces I thought were unlikely to be impacted and that I just had general questions on.

cvchaparro:
> @cvchaparro is this PR ready for full review? You had sent a status yesterday about wanting to do some reworks, but I haven't seen a commit around that. So, I just hit pieces I thought were unlikely to be impacted and that I just had general questions on.

The changes are in, now. Should be ready for official review.

python/src/aac/plugins/validators/__init__.py
cvchaparro: Using the same Python LSP server could be beneficial but "standardizing on implementation and tooling" is not something I'm sure we should do (depending on how "standard" you want it to be - i.e. do you mean everyone has to use the same editor/IDE?).

For the record, I just tested renaming the `OperationCancelled` class in VSCode using Jedi and Pylance and it also updates the string in `__all__` so even standardizing LSP server doesn't seem totally necessary unless there are *other* reasons to do so.

lizzcondrey: It's definitely something I am back and forth on because there are pretty even pros and cons to standardizing and to being diverse too. Primarily thinking from it in an onboarding stance of "here's our base tooling configuration" that can be provided to a new dev to get them going.

python/src/aac/plugins/validators/_validator_findings.py
```diff
@@ -18,9 +19,10 @@ class ValidatorFindings:
 
     findings: list[ValidatorFinding] = attrib(default=Factory(list), validator=validators.instance_of(list))
 
-    def add_findings(self, new_findings: list[ValidatorFinding]) -> None:
+    def add_findings(self, new_findings: Union[list[ValidatorFinding], "ValidatorFindings"]) -> None:
```
lizzcondrey, line 22: Not following what making new findings a union of the list of findings and the string identifying as findings gains?

python/src/aac/plugins/validators/_validator_findings.py
```diff
@@ -18,9 +19,10 @@ class ValidatorFindings:
...
         """Add the new findings to the collection of all findings."""
-        self.findings.extend(new_findings)
+        findings = new_findings if isinstance(new_findings, list) else new_findings.get_all_findings()
```
lizzcondrey, line 24: Like the additional sanity check of the `isinstance` for verifying that what's passed in truly is a list as desired. 

And the else gets you a list if you pass what would be a higher object that holds the findings, if I am following correctly?

lizzcondrey:
Not finished with checking this, needing to look again over the validator impl and yaml with fresh eyes tomorrow. I was confusing myself earlier looking it over while trying to get some reporting stuff sorted. Sorry for the delay.

python/src/aac/plugins/validators/_validator_findings.py
cvchaparro: This is just a type hint. The `Union` is Python's way of showing that the type of an object can be one of multiple things - in this case a list of `ValidatorFinding` objects or a `ValidatorFindings` object (the string is used here because in the definition of the class, Python still doesn't know what `"ValidatorFindings"` is so I think it's akin to a forward declaration in C++, or something). Basically it allows us to (using Java terms) overload the function for either type of parameter but (to my knowledge) Python doesn't allow overloading the way it's done in statically-typed languages.

The reason I'm doing this is so that we can either add findings from a list of findings (the original behavior) or get the findings from the passed in `ValidatorFindings` object to combine them (imagine overloading the `+` operator for the `ValidatorFindings` object) which allows me to extract the "add finding" part into a function and combine all the collected findings together.

python/src/aac/plugins/validators/_validator_findings.py
cvchaparro: Given that we're claiming to accept the two types of objects described in [this comment](https://github.com/DevOps-MBSE/AaC/pull/667#discussion_r1320466604) the expectation is that if we get a `list` we just use that whereas if we get anything else we assume it's a `ValidatorFindings` object which exposes a `get_all_findings()` method. As you allude to, passing in anything else would break things.

python/src/aac/plugins/validators/_validator_findings.py
lizzcondrey: Thanks for the thorough explanation!

python/src/aac/plugins/validators/_validator_findings.py
lizzcondrey: Would it be useful to wrap that else in a try/catch since if it is something other than the two types of findings objects that `get_all_findings()` wouldn't exist and we could then handle the break more gracefully? Or is there a higher level handling of that break that I missed/am unaware of?

Coffee2Bits (approved):
Looks good! It even caught an issue in the AlarmClock model usecase!

```
Validation ERROR from 'Usecase sources and targets refer to defined participants' in /workspace/AaC/python/model/alarm_clock/usecase.yaml. Source 'alarm' of step 'The alarm is activated to emit a noise from the AlarmClock speaker' does not refer to a participant of the usecase.

Validation ERROR from 'Usecase sources and targets refer to defined participants' in /workspace/AaC/python/model/alarm_clock/usecase.yaml. Target 'alarm' of step 'The timer emits an alarm notification to the alarm' does not refer to a participant of the usecase.
```

lizzcondrey:
> Looks good! It even caught an issue in the AlarmClock model usecase!
> 
> ```
> Validation ERROR from 'Usecase sources and targets refer to defined participants' in /workspace/AaC/python/model/alarm_clock/usecase.yaml. Source 'alarm' of step 'The alarm is activated to emit a noise from the AlarmClock speaker' does not refer to a participant of the usecase.
> 
> Validation ERROR from 'Usecase sources and targets refer to defined participants' in /workspace/AaC/python/model/alarm_clock/usecase.yaml. Target 'alarm' of step 'The timer emits an alarm notification to the alarm' does not refer to a participant of the usecase.
> ```
> 
> Edit - I approved before realizing that Lizz has outstanding comments and review.

All good, I just finished looking it over and was about to conditionally approve with the last question I had about the modification of the `ValidatorFindings` and potentially adding some additional error handling.

Agree looks good and you did what I was about to with using it on one of the example models.

python/src/aac/plugins/validators/__init__.py
cvchaparro: The word "tooling" is extremely general - can you please give some examples or more specific description of which specific tools you would potentially like to standardize?

python/src/aac/plugins/validators/_validator_findings.py
cvchaparro: It will probably slow some things down since the checks would happen at runtime but if we want to do this, in general, I'd prefer to use [`typeguard`](https://typeguard.readthedocs.io/en/latest/userguide.html) over polluting the project with a bunch of `try`/`catch` blocks. We actually already use the `check_type` function from the `typeguard` library in a couple places, so it shouldn't be crazy to add it but I don't know the performance penalty we'll get by adding that everywhere (and it probably makes sense to do it everywhere or in very few places/nowhere, IMO).

cvchaparro:
> > Looks good! It even caught an issue in the AlarmClock model usecase!
> > ```
> > Validation ERROR from 'Usecase sources and targets refer to defined participants' in /workspace/AaC/python/model/alarm_clock/usecase.yaml. Source 'alarm' of step 'The alarm is activated to emit a noise from the AlarmClock speaker' does not refer to a participant of the usecase.
> > 
> > Validation ERROR from 'Usecase sources and targets refer to defined participants' in /workspace/AaC/python/model/alarm_clock/usecase.yaml. Target 'alarm' of step 'The timer emits an alarm notification to the alarm' does not refer to a participant of the usecase.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Edit - I approved before realizing that Lizz has outstanding comments and review.
> 
> All good, I just finished looking it over and was about to conditionally approve with the last question I had about the modification of the `ValidatorFindings` and potentially adding some additional error handling.

To which modification of the `ValidatorFindings` are you referring - the `try`/`catch` block thing? If so, what additional error handling are you looking for?

> 
> Agree looks good and you did what I was about to with using it on one of the example models.

python/src/aac/plugins/validators/_validator_findings.py
lizzcondrey: So, quick skim of the link I was about to ask what's the difference between it and `isintance` and it has a line stating it's "a beefed up version". 

The `typeguard` package has a [runtime component](https://typeguard.readthedocs.io/en/latest/userguide.html#using-the-import-hook) that I think helps address my confusion over the `Union` modification you made for `ValidatorFindings` and wouldn't make us have to use the string name reference. 

I lean towards wanting to have that additional type check because of the number of layers that seemed to be involved with validation.

lizzcondrey:
> > > All good, I just finished looking it over and was about to conditionally approve with the last question I had about the modification of the `ValidatorFindings` and potentially adding some additional error handling.
> > 
> > 
> > To which modification of the `ValidatorFindings` are you referring - the `try`/`catch` block thing? If so, what additional error handling are you looking for?
> > > Agree looks good and you did what I was about to with using it on one of the example models.
> 
> @lizzcondrey did you want to see other changes or did I misunderstand your last comment?

I was wanting to finish the discussion around the `try`/`catch` and addressing the breakage for if an object of neither finding type comes in since we realize already that would cause it to break.

python/src/aac/plugins/validators/__init__.py
lizzcondrey: So, don't think it makes sense to "mandate" the IDE that is used, outside of some work with VSCode because of the extension we develop. 

But having a list of "quality of life" type tools could be useful, i.e. LSP server, linters, static analysis, etc. So, I am meaning "tools" more akin to IDE extensions. 

Again, primarily from the onboarding side of things to know what a new dev needs to get installed. But also help limit introducing odd environment behaviors and get the lovely line of "Well it runs on my machine". 

Thankfully there isn't a lot of additional environment installations that are part of being able to use AaC like I've had on previous projects to potentially introduce rogue behavior. So, it might not really be necessary to hard analyze and work at coming up with a "tooling" list. And we have the environment configuration files for GitPod and Emacs that help address this in some ways (at least I know GitPod does). 

Based on the pieces of my comment above, I may just be overthinking this and worrying about unlikely impacts. Which is quite plausible.

python/src/aac/plugins/validators/_validator_findings.py
cvchaparro: I think it'd make sense to try that but I don't think it's in scope for this story.

python/src/aac/plugins/validators/__init__.py
cvchaparro: I'd say Python-specific tools like that (minus the Python LSP server) are included in the dev dependencies in `setup.py` already. And VSCode extension dev tools should be in the `package.json` file. If we find more tools we need to have people use, we should just add them there, IMO. That said, I don't see an action that needs to be taken, now, with respect to this.

python/src/aac/plugins/validators/_validator_findings.py
lizzcondrey: Okay, I'll generate a backlog item for further evaluating `typeguard` and determining implementation needs.

lizzcondrey: Here is the [issue](https://github.com/DevOps-MBSE/AaC/issues/673) associated with this discussion.

PR #708, opened 2023-10-20
Print Validated File List

### LINKED ISSUE #571: I want the validation success message to list all files that were validated
As an AaC user, I want the validation success message to list all files that were validated, so that I can have a better understanding of which files are being pulled in to the validation process.

## AC:
- [ ] On success, the validation message reports the whole set of file uris found in the validated definitions.

lizzcondrey:
# Description
Added listing definitions in which files that have been validated. 

# Linked Items:
Closes #571 

### Added
- When `_validate` makes a check for being valid, it prints out the definitions within files that have been validated. 

# Checklist:
- [x] I updated project documentation to reflect my changes.
- [x] I bumped the version.

### DIFF
--- a/python/src/aac/plugins/validators/_validator_result.py
+++ b/python/src/aac/plugins/validators/_validator_result.py
@@ -45,6 +45,6 @@ class ValidatorResult:
                 if definition.source.uri not in definition_sources:
                     definition_sources.append(definition.source.uri)
 
-            valid_files_output = "\n".join(f"The definitions in {definition_file}" for definition_file in definition_sources)
+            valid_files_output = "\nThe following definition files have been validated:\n" + "\n".join(definition_file for definition_file in definition_sources)
 
-        return valid_files_output + "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
+        return valid_files_output + "\n" + "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
--- a/python/src/aac/plugins/validators/_validator_result.py
+++ b/python/src/aac/plugins/validators/_validator_result.py
@@ -36,5 +36,9 @@ class ValidatorResult:
                 f"\nValidation {finding.severity.name} from '{loc.validation_name}' in {loc.source.uri} "
                 f"at {loc.location.line + 1}:{loc.location.column} {finding.message}"
             )
+        
+        if len(self.findings.get_error_findings()) == 0:
+            validated_defintions = self.definitions
+            return "\n".join(f"The definitions in {definition.source.uri}" for definition in validated_defintions)
 
         return "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
--- a/python/src/aac/plugins/validators/_validator_result.py
+++ b/python/src/aac/plugins/validators/_validator_result.py
@@ -43,7 +43,11 @@ class ValidatorResult:
                 if definition.source.uri not in definition_sources:
                     definition_sources.append(definition.source.uri)
 
+<<<<<<< Updated upstream
             return "\n".join(f"The definitions in {definition_file} meet all validation constraints." for definition_file in definition_sources)
+=======
+            return "\n".join(f"The definitions in {definition.source.uri}" for definition in definition_sources)
+>>>>>>> Stashed changes
 
         else:
             return "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
--- a/python/src/aac/plugins/validators/_validator_result.py
+++ b/python/src/aac/plugins/validators/_validator_result.py
@@ -33,7 +33,8 @@ class ValidatorResult:
         def format_message(finding: ValidatorFinding) -> str:
             loc = finding.location
             return (
-                f"\nValidation {finding.severity.name} from '{loc.validation_name}' in {loc.source.uri}. {finding.message}"
+                f"\nValidation {finding.severity.name} from '{loc.validation_name}' in {loc.source.uri} "
+                f"at {loc.location.line + 1}:{loc.location.column} {finding.message}"
             )
 
         return "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
--- a/python/src/aac/plugins/validators/_validator_result.py
+++ b/python/src/aac/plugins/validators/_validator_result.py
@@ -43,7 +43,7 @@ class ValidatorResult:
                 if definition.source.uri not in definition_sources:
                     definition_sources.append(definition.source.uri)
 
-            return "\n".join(f"The definitions in {definition.source.uri}" for definition in definition_sources)
+            return "\n".join(f"The definitions in {definition_file}" for definition_file in definition_sources)
 
         else:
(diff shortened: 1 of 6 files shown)


python/src/aac/validate/_validate.py
```diff
@@ -78,6 +78,9 @@ def _with_validation(
         result = _validate_definitions(user_definitions, language_context, validate_context)
 
         if result.is_valid():
+            print("The following file(s) have been validated:")
```
lizzcondrey, line 81: While I am fairly confident this gets us the information we wanted with this task, I am not sure that just printing is what was desired.

python/src/aac/validate/_validate.py
```diff
@@ -78,6 +78,8 @@ def _with_validation(
         result = _validate_definitions(user_definitions, language_context, validate_context)
 
         if result.is_valid():
+            for definition in user_definitions:
+                print(f"The {definition.name} definition in {definition.source.uri} has been validated.")
```
lizzcondrey, line 82: While I am fairly confident this gets us the information we wanted with this task, I am not sure that just printing is what was desired.

cvchaparro: You are correct - this isn't what we want since it would bypass `click` which, ultimately, should handle all of our output.

What we really want is to update the `get_messages_as_string` function.
https://github.com/DevOps-MBSE/AaC/blob/ebeb7c1831ac3586b0932a7ed30e2f061af14519/python/src/aac/plugins/validators/_validator_result.py#L30-L40

and add an `if`-check to ensure we only add the files that were processed if validation was successful but that class will already have all the data you need.

cvchaparro: Finally, I think it will be too noisy to print out every definition name in addition to every file name. We should limit the extra output to just filenames.

cvchaparro (changes requested):
As you pointed out, that's not really the desired end-state. That said, I pointed you to the location in which these changes should be made.

python/src/aac/validate/_validate.py
lizzcondrey: Okay, I think I got this cleaned up. Please give it another look.

python/src/aac/plugins/validators/_validator_result.py
```diff
@@ -37,4 +37,13 @@ def format_message(finding: ValidatorFinding) -> str:
                 f"at {loc.location.line + 1}:{loc.location.column} {finding.message}"
             )
 
-        return "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
+        if len(self.findings.get_error_findings()) == 0:
```
cvchaparro, line 40: We shouldn't replicate the check, here:

(See below for reasoning related to `valid_files_output`)

```suggestion
        valid_files_output = ""
        if self.is_valid():
```

python/src/aac/plugins/validators/_validator_result.py
```diff
@@ -37,4 +37,13 @@ def format_message(finding: ValidatorFinding) -> str:
...
+            definition_sources = []
+            for definition in self.definitions:
+                if definition.source.uri not in definition_sources:
+                    definition_sources.append(definition.source.uri)
+
+            return "\n".join(f"The definitions in {definition_file}" for definition_file in definition_sources)
+
+        else:
+            return "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
```
cvchaparro, line 49: This will cause us to lose information if the definitions are valid but, for example, `WARNING` or `INFO` findings were added to the result.

I can't guarantee this is exactly the code we want here but I think something like this would work (though haven't actually tested it to see). The goal is to show the validated files (assuming there were no `ERROR` messages) in addition to all the `WARNING` and `INFO` messages:

```suggestion
            valid_files_output = "\n".join(f"{definition_file} is valid" for definition_file in definition_sources)

        return valid_files_output + "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
```

cvchaparro (changes requested):
Almost there

python/src/aac/plugins/validators/_validator_result.py
```diff
@@ -37,4 +37,14 @@ def format_message(finding: ValidatorFinding) -> str:
...
-        return "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
+        valid_files_output = ""
+
+        if self.is_valid():
+            definition_sources = []
+            for definition in self.definitions:
+                if definition.source.uri not in definition_sources:
+                    definition_sources.append(definition.source.uri)
+
+            valid_files_output = "\n".join(f"The definitions in {definition_file}" for definition_file in definition_sources)
```
cvchaparro, line 48:
```suggestion
            valid_files_output = "\n".join(f"{definition_file} is valid." for definition_file in definition_sources)
```

lizzcondrey:
Alright, think maybe got this cleaned up better now... Here is what I've got the output set to look like.

```
(venv) gitpod /workspace/AaC/python/model/alarm_clock (lizzcondrey/571-i-want-the-validation-success-message-to-list-all-files-that-were-validated) $ aac validate alarm_clock.yaml 
alarm_clock.yaml is valid.

The following definition files have been validated:
/workspace/AaC/python/model/alarm_clock/alarm_clock.yaml
/workspace/AaC/python/model/alarm_clock/structures.yaml

Validation INFO from 'Validate definition references' in /workspace/AaC/python/src/aac/plugins/validators/unique_requirement_ids/unique_requirement_ids.yaml at 2:8 No references to 'Validate requirement IDs are not duplicated.' in the language context.
```

python/src/aac/plugins/validators/_validator_result.py
```diff
@@ -37,4 +37,14 @@ def format_message(finding: ValidatorFinding) -> str:
...
-        return "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
+        valid_files_output = ""
+
+        if self.is_valid():
+            definition_sources = []
+            for definition in self.definitions:
+                if definition.source.uri not in definition_sources:
+                    definition_sources.append(definition.source.uri)
+
+            valid_files_output = "\nThe following definition files have been validated:\n" + "\n".join(definition_file for definition_file in definition_sources)
```
cvchaparro, line 48:
```suggestion
            valid_files_output = "\nThe following additional definition files have been validated:\n" + "\n".join(definition_file for definition_file in definition_sources)
```

python/src/aac/plugins/validators/_validator_result.py
```diff
@@ -37,4 +37,14 @@ def format_message(finding: ValidatorFinding) -> str:
...
+
+        return valid_files_output + "\n" + "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
```
cvchaparro, line 50:
```suggestion
        return f"{valid_files_output}\n" + "\n".join([format_message(finding) for finding in self.findings.get_all_findings()])
```

cvchaparro (changes requested):
Two small items but I think they'll make things more clear/clean.

cvchaparro (approved):
LGTM

PR #866, opened 2024-09-05
Migrate Build Chain in Core AaC Repo

### LINKED ISSUE #840: 📋 [TASK] - Migrate Build Chain in Core AaC Repo
### Parent Feature

#838

### Task Description

As an AaC Developer, I want to update the core AaC repository to use the new build chain process, so that I have a consistent build chain across project repositories.

### Acceptance Criteria

- [ ] Replace setup.py file with a pyproject.toml file in core AaC repo.
- [ ] Edit other build and infrastructure files as needed.
  - [ ] Manifest.in
  - [ ] tox.ini
  - [ ] README
- [ ] Pull latest build workflows from organization templates as needed

### Unresolved questions

_No response_

bunchw:
# Description

Updating AaC to use the new build chain process.

# Linked Items:

Closes/Fixes/Resolves #840 

### Added

Added a pyproject.toml file, and a MANIFEST.in file

### Changed

### Deprecated

### Removed

### Fixed

- _Describe any bug fixes._

### Security

# Checklist:

- [ ] I updated new and existing unit tests to account for my changes.
- [x] I bumped the version.


README.md
```diff
@@ -164,3 +164,39 @@ Contributors, developers, or just generally interested parties who would like to
...
+
+Previously, this project was built with dependency information kept in a setup.py script.
+However, the preferred method is to use pyproject.toml to set the project-level options.
+Required modules are kept in the dependency sections of the pyproject.toml, and then the
+pip-compile command is used to add hashes to the requirements.txt file for enhanced security
+(see additional instructions below).
+
+To coincide with these changes, some changes to tox.ini and the addition of a MANIFEST.ini file were also necessary.
+
+These lines were added to tox.ini:
```
lizzcondrey, line 178: Remember to add the block quotes for the code snippets.

python/pyproject.toml
```diff
@@ -1,18 +1,28 @@
...
-
-try:
-    with open(README_FILE_PATH, "r") as file_handler:
-        readme_description = file_handler.read()
-except Exception as exception:
-    logging.error(f"Failed to open readme file {README_FILE_PATH} with error:{linesep}  {exception}")
-    readme_description = ""
+[project]
+name = "aac"
+version =  "0.4.18"
```
lizzcondrey, line 7: Thought we were on 0.4.22? But should be able to pull the version in a similar way here that was being done in the `setup.py` from the `src/aac/__init__.py` file. Double check. It should be making use of the dynamic flagging to be able to do so if I remember correctly. That will allow for the version plugin to remain operating as is. Otherwise, we will need to do some rework of the plugin as well.

python/pyproject.toml
```diff
@@ -1,18 +1,28 @@
...
+authors = [
+    { name="bunchw", email="asdfasdaf@email.com" },
+    { name="lizzcondrey", email="asdfasdfd@mail.com" },
```
lizzcondrey, line 10: Remember we were wanting to try and shift this to being the team label instead so that we don't have to keep adding team members names.

python/secure_install/pipeline_scripts/create_install_archive.bash
```diff
@@ -18,7 +18,7 @@ cp README.md $install_dir
 cd $install_dir
 
 pip wheel ../../
-pip-compile --generate-hashes ../../setup.py
+pip-compile --generate-hashes pyproject.toml
```
lizzcondrey, line 21: Still located same place, so shouldn't drop the drill down for the file path, I don't believe.

lizzcondrey: Unrelated, and definitely confusing me, but seems we've missed changing this in the other repos, so not sure how it's gone through fine in pipeline....Wooo

python/tests/test_aac/context/test_language_context.py
```diff
@@ -46,7 +46,7 @@ def test_get_definitions_by_name(self):
 
         context.parse_and_load(VALID_AAC_YAML_CONTENT_SPACE_IN_NAME)
         definitions = context.get_definitions_by_name("Test Schema")
-        self.assertEqual(definitions[0].name, "Test Schema")
+        self.assertEqual(definitions[0].name.replace(" ", ""), "TestSchema")
```
lizzcondrey, line 49: Interesting and again confused, you did a bug adjustment for allowing spaces in definition names, so not sure why this had to be altered?

lizzcondrey (changes requested):

python/tests/test_aac/context/test_language_context.py
bunchw: Because the names are retrieved in python form, which cannot have spaces.

python/pyproject.toml
```diff
@@ -1,18 +1,27 @@
...
-
-try:
-    with open(README_FILE_PATH, "r") as file_handler:
-        readme_description = file_handler.read()
-except Exception as exception:
-    logging.error(f"Failed to open readme file {README_FILE_PATH} with error:{linesep}  {exception}")
-    readme_description = ""
+[project]
+name = "aac"
+dynamic =  ["version"]
```
lizzcondrey, line 7: I'm not seeing where you tell this to look for pulling the version from elsewhere in the codebase?

python/tests/test_aac/context/test_language_context.py
lizzcondrey: But then how is this test not breaking till now? The bug was addressing being able to have spaces and find them accordingly I had thought?

bunchw: I am not sure why its only now breaking, but I do remember having to do a similar fix for another test.  There is no way to have the method get_definitions_by_name return a name with a space in it.  You can give the method a space, but it will always return a dictionary where the name has no space, because it uses python names.

lizzcondrey: Yeah, the bug you addressed about spaces in the names is what allowed for that line to be what is was previously. Not at all understanding why it would need to be changed now. 

https://github.com/DevOps-MBSE/AaC/pull/863/files

python/pyproject.toml
bunchw: You are right, thank you for catching that, I added the table for dynamic variables.  I suspect theres a simpler way to do it, but this is the only method I've found that works for the setup we have right now.

lizzcondrey: I wonder if this could be part of what was causing the pipeline spaz out? intrigued to see

python/tests/test_aac/context/test_language_context.py
lizzcondrey: > I am not sure why its only now breaking, but I do remember having to do a similar fix for another test. There is no way to have the method get_definitions_by_name return a name with a space in it. You can give the method a space, but it will always return a dictionary where the name has no space, because it uses python names.

So a similar check is made with the manual removal in the test for the definition parser, so was going to originally say that this shouldn't have to change then. But, that PR I linked introduced that check and the space handling in language context itself, and it passed then as a result. So, definitely not understanding what's happened...

lizzcondrey: Gaaah, love when the editor decides to be funky about how it wants to categorize the line -.-

So, opposing the point of the test does not meet our needs. That test set was explicitly around evaluating can the name be found with  space in it. I think what it has now highlighted is that our way of addressing the spaces in a name may have been too simple/an incorrect approach. 

I was able to get this test to continue passing by tweaking the referenced definition name to be "Test Schema2" since "TestSchema" exists because of the other provided definition for testing. That let the individual test pass and the test suite of TestLanguageContext. I think that is the right way to tweak this test for right now to get it to pass within the pipeline and we need to flag a bug to take another look at handling names with spaces.

bunchw: Huh, I also tried that and now it is working in both places.  I am confused on what is happening here, I guess I missed something or misdiagnosed the issue.  It looks like it's fine now with the spaces included, I'm not sure why it seems so inconsistent.

lizzcondrey: So we hit something similar to this in legacy to 0.4.0 about the context being odd to persist between tests sometimes. I think that was contributing somewhat, but I also think that since we added the handling to permitting spaces it was causing a name conflict with two `TestSchema` definitions. So, we need to file a bug or add a note on the associated test suite task to make suer to look into it more thoroughly. Once that is done, believe this PR is good to merge. Since you initially encountered the issue, would like you to file the bug or post a comment to the test suite issue, please.

lizzcondrey (dismissed):

python/tests/test_aac/context/test_language_context.py
bunchw: Bug is filed on Jira.

lizzcondrey (approved):

PR #883, opened 2024-10-15
Improve Tests in Execute Unit Tests

### LINKED ISSUE #850: 📋 [TASK] - Improve Tests in Execute Unit Tests
### Parent Feature

#844

### Task Description

As an AaC Test Developer, I want to expand the test coverage in my unit tests, so that I can have more confidence in the robustness of my code.

### Acceptance Criteria

- [ ] Update test logic as necessary to expand test coverage. 
  - [ ] Happy path and sad path at a minimum
- [ ] (Stretch) Incorporate a behave feature file for testing if applicable.
- [ ] Run `flake8` on the targeted test directory and address findings.
- [ ] Document any identified gaps.

### Unresolved questions

_No response_

bunchw:
# Description

Improve Tests in Execute Unit Tests.

# Linked Items:

Closes/Fixes/Resolves #850 

### Added

ExecutionResult test class.

### Changed

### Deprecated

### Removed

### Fixed

- _Describe any bug fixes._

### Security

# Checklist:

- [x] I bumped the version.

### DIFF
--- a/python/src/aac/execute/aac_execution_result.py
+++ b/python/src/aac/execute/aac_execution_result.py
@@ -73,10 +73,15 @@ class ExecutionResult:
 
     def add_message(self, message: ExecutionMessage) -> None:
         """Add a message to the list of messages."""
+        if type(message) is not ExecutionMessage:
+            raise TypeError("Message should be of class ExecutionMessage")
         self.messages.append(message)
 
     def add_messages(self, messages: list[ExecutionMessage]) -> None:
         """Add messages to the list of messages."""
+        for message in messages:
+            if type(message) is not ExecutionMessage:
+                raise TypeError("Message should be of class ExecutionMessage")
         self.messages.extend(messages)
 
     def is_success(self) -> bool:
@@ -87,15 +92,12 @@ class ExecutionResult:
         """Return the output messages as a combined string."""
         result = ""
         for message in self.messages:
-            if type(message) is ExecutionMessage:
-                result += message.message + linesep
-                if message.source is not None:
-                    result += f"  Source: {message.source.uri}"
-                    if message.location is not None:
-                        result += f" (Ln {message.location.line}: Col {message.location.column}: Pos {message.location.position}: Spn {message.location.span})"
-                    result += linesep
-            else:
-                result += message + linesep
+            result += message.message + linesep
+            if message.source is not None:
+                result += f"  Source: {message.source.uri}"
+                if message.location is not None:
+                    result += f" (Ln {message.location.line}: Col {message.location.column}: Pos {message.location.position}: Spn {message.location.span})"
+                result += linesep
         return result
 
 
--- a/python/src/aac/execute/aac_execution_result.py
+++ b/python/src/aac/execute/aac_execution_result.py
@@ -87,12 +87,15 @@ class ExecutionResult:
         """Return the output messages as a combined string."""
         result = ""
         for message in self.messages:
-            result += message.message + linesep
-            if message.source is not None:
-                result += f"  Source: {message.source.uri}"
-                if message.location is not None:
-                    result += f" (Ln {message.location.line}: Col {message.location.column}: Pos {message.location.position}: Spn {message.location.span})"
-                result += linesep
+            if type(message) == ExecutionMessage:
+                result += message.message + linesep
+                if message.source is not None:
+                    result += f"  Source: {message.source.uri}"
+                    if message.location is not None:
+                        result += f" (Ln {message.location.line}: Col {message.location.column}: Pos {message.location.position}: Spn {message.location.span})"
+                    result += linesep
+            else:
+                result += message + linesep
         return result
 
 
--- a/python/src/aac/execute/aac_execution_result.py
+++ b/python/src/aac/execute/aac_execution_result.py
@@ -87,7 +87,7 @@ class ExecutionResult:
         """Return the output messages as a combined string."""
         result = ""
         for message in self.messages:
-            if type(message) == ExecutionMessage:
+            if type(message) is ExecutionMessage:
                 result += message.message + linesep
                 if message.source is not None:
                     result += f"  Source: {message.source.uri}"
--- a/python/tests/test_aac/execute/test_aac_execution_result.py
+++ b/python/tests/test_aac/execute/test_aac_execution_result.py
@@ -1,0 +1,24 @@
+import unittest
+from aac.execute.aac_execution_result import ExecutionResult, ExecutionMessage, ExecutionStatus, MessageLevel
+
+class TestExecitionResult(unittest.TestCase):
+
+    def test_execution_result(self):
+        execution_result = ExecutionResult("name", "command_name", ExecutionStatus.GENERAL_FAILURE, [])
+        execution_result.add_message("message 1")
+        execution_result.add_messages([
+            ExecutionMessage("message 2", MessageLevel.INFO, None, None),
+            ExecutionMessage("message 3", MessageLevel.INFO, None, None),
+            ])
+        execution_result.status_code = ExecutionStatus.SUCCESS
+
+        messages = execution_result.get_messages_as_string()
+        print(messages)
+        self.assertEqual(messages, "message 1\nmessage 2\nmessage 3\n")
+        self.assertEqual(execution_result.status_code, ExecutionStatus.SUCCESS)
+
+    def test_execution_result_no_message(self):
+        with self.assertRaises(TypeError):
+            execution_result = ExecutionResult("name", "command_name", None, None)
+
+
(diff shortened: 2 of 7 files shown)


lizzcondrey (changes requested):
Missing exercising of the actual `PluginRunner` class in the `test_plugin_runner.py`
Missing exercising the `PluginManager` class, so should have a `test_plugin_manager.py`

Following that same thread, missing exercising the `command_line.py`, so should have a `test_command_line.py` file.  But less certain it is needed since all of our plugin unit tests have to call the `initialize_cli` which relies on all of those cli methods to be able to execute for the test.

python/src/aac/execute/aac_execution_result.py
```diff
@@ -87,12 +87,15 @@ def get_messages_as_string(self) -> str:
...
-                result += linesep
+            if type(message) is ExecutionMessage:
+                result += message.message + linesep
+                if message.source is not None:
+                    result += f"  Source: {message.source.uri}"
+                    if message.location is not None:
+                        result += f" (Ln {message.location.line}: Col {message.location.column}: Pos {message.location.position}: Spn {message.location.span})"
+                    result += linesep
+            else:
+                result += message + linesep
```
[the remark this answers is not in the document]
bunchw: It seemed like the purpose was to be able to pass in strings as messages as well, which would cause this method to fail.  Is that not the intended functionality?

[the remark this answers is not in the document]
lizzcondrey: So, strings can be added to messages at lower levels, but once it has made it up to here, it should have been put in the proper format of being an `ExcecutionMessage` object. If it has not been properly formatted before now, then that is a gap in the data handling. I believe we would want it to fail, but can have handling for that failure of returning a notice to the command line of something like "Internal handling error of messages not being properly formatted"

python/tests/test_aac/execute/test_aac_execution_result.py
```diff
@@ -1,24 +1,47 @@
-import unittest
+from unittest import TestCase
 from aac.execute.aac_execution_result import ExecutionResult, ExecutionMessage, ExecutionStatus, MessageLevel
 
-class TestExecitionResult(unittest.TestCase):
+class TestExecitionResult(TestCase):
```
lizzcondrey, line 4:
```suggestion
class TestExecutionResult(TestCase):
```

python/tests/test_aac/execute/test_aac_execution_result.py
```diff
@@ -1,24 +1,47 @@
...
+
+        self.assertEqual(execution_success.status_code, ExecutionStatus.SUCCESS)
+        self.assertEqual(execution_constraint_failure.status_code, ExecutionStatus.CONSTRAINT_FAILURE)
+        self.assertEqual(execution_constraint_warning.status_code, ExecutionStatus.CONSTRAINT_WARNING)
+        self.assertEqual(execution_parser_failure.status_code, ExecutionStatus.PARSER_FAILURE)
+        self.assertEqual(execution_plugin_failure.status_code, ExecutionStatus.PLUGIN_FAILURE)
+        self.assertEqual(execution_operation_cancelled.status_code, ExecutionStatus.OPERATION_CANCELLED)
+        self.assertEqual(execution_general_failure.status_code, ExecutionStatus.GENERAL_FAILURE)
 
     def test_execution_result_no_message(self):
```
lizzcondrey, line 42: Should be able to alter this to be more encompassing of 
```suggestion
    def test_execution_result_type_errors(self):
```
And do checks for if you provide improper entries for any of the `ExecutionResult` attributes since they are all typed (as much as you can in Python) and have their `instance_of` checks for being expected type coming in.

lizzcondrey (changes requested):

python/tests/test_aac/execute/test_aac_execution_result.py
```diff
@@ -0,0 +1,47 @@
+from unittest import TestCase
+from aac.execute.aac_execution_result import ExecutionResult, ExecutionMessage, ExecutionStatus, MessageLevel
+
+class TestExecitionResult(TestCase):
+
+    def test_execution_result(self):
```
lizzcondrey, line 6: I would break this apart into multiple tests.

One can be checking that all messages exist as expected.

Another can be checking that all `MessageLevels` are set as expected.

Another would be checking that name was set as expected.

Another would be checking that the command name was set as expected. 

And like in another comment, testing when invalid is provided. Similar to what/how the plugin runner was exercised.

lizzcondrey (changes requested):

python/tests/test_aac/execute/test_aac_execution_result.py
```diff
@@ -0,0 +1,47 @@
...
+        execution_result = ExecutionResult("name", "command_name", ExecutionStatus.GENERAL_FAILURE, [])
+        execution_result.add_messages([
+            ExecutionMessage("message 1", MessageLevel.DEBUG, None, None),
+            ExecutionMessage("message 2", MessageLevel.INFO, None, None),
+            ExecutionMessage("message 3", MessageLevel.WARNING, None, None),
+            ExecutionMessage("message 4", MessageLevel.ERROR, None, None)
+            ])
+        execution_result.status_code = ExecutionStatus.SUCCESS
```
lizzcondrey, line 14: And since you change the status here, can add a check prior for evaluating the full `is_success` logic.
```suggestion
        self.assertFalse(execution_result.is_success)
        
        execution_result.status_code = ExecutionStatus.SUCCESS
```

lizzcondrey (changes requested):

python/src/aac/execute/aac_execution_result.py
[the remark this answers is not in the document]
bunchw: I made it return an error if a message that isnt an ExecutionMessage is passed through.  Is that good?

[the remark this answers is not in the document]
lizzcondrey: Yeap, I had seen that modification and the tweak to the test to reflect.

python/tests/test_aac/execute/test_aac_execution_result.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(execution_constraint_failure.status_code, ExecutionStatus.CONSTRAINT_FAILURE)
+        self.assertEqual(execution_constraint_warning.status_code, ExecutionStatus.CONSTRAINT_WARNING)
+        self.assertEqual(execution_parser_failure.status_code, ExecutionStatus.PARSER_FAILURE)
+        self.assertEqual(execution_plugin_failure.status_code, ExecutionStatus.PLUGIN_FAILURE)
+        self.assertEqual(execution_operation_cancelled.status_code, ExecutionStatus.OPERATION_CANCELLED)
+        self.assertEqual(execution_general_failure.status_code, ExecutionStatus.GENERAL_FAILURE)
+
+    def test_execution_result_invalid_status(self):
+        with self.assertRaises(AttributeError):
+            execution_result = ExecutionResult("name", "command_name", ExecutionStatus.general_failure, [])
```
lizzcondrey, line 62: Can have an additional test case of not even using the enum and be sure that is handled.

```suggestion
        with self.assertRaises(AttributeError):
            execution_result = ExecutionResult("name", "command_name", ExecutionStatus.general_failure, [])
            
        with self.assertRaises(AttributeError):
            execution_result = ExecutionResult("name", "command_name", "general_failure", [])
```
So, something like this.

python/tests/test_aac/execute/test_aac_execution_result.py
lizzcondrey: For the name entries, it would be provide something other than a string, so similar to the invalid tests for the `AaCCommand` and `AaCArgument` in the `plugin_runner` tests

python/tests/test_aac/execute/test_aac_execution_result.py
lizzcondrey: And you will do a similar test set for `MessageLevel`

lizzcondrey (changes requested):

lizzcondrey (changes requested):

python/tests/test_aac/execute/test_plugin_manager.py
```diff
@@ -0,0 +1,25 @@
...
+
+        plugin_manager.unregister(plugin)
+        plugins = plugin_manager.get_plugins()
+        self.assertNotIn(str(plugin), ''.join(str(plugins)))
+
+        plugin_manager.register(plugin)
+        plugins = plugin_manager.get_plugins()
+        self.assertTrue(str(plugin), ''.join(str(plugins)))
+
+    def test_plugin_manager_fail(self):
```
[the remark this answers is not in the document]
bunchw: So, for some reason when using python -m unittest pluggy will accept non module type arguments.  When using pytest it does not and returns an error.  The pytest functionality seems like it should be the correct one, but this isnt something we can change as it is apart of the pluggy dependency.  Without this test, I really cant think of a way to test a fail case for plugin manager, its a very small file and I dont think users would ever realistically use it directly.  If it can accept anything, then it's pretty much foolproof.

lizzcondrey (changes requested):

python/tests/test_aac/execute/test_plugin_manager.py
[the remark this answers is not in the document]
lizzcondrey: Decided to incorporate a comment explaining the current issue implementing more of a failure state test to do investigations at a later time.

PR #886, opened 2024-10-17
Improve Tests in Context Unit Tests

### LINKED ISSUE #853: 📋 [TASK] - Improve Tests in Context Unit Tests
### Parent Feature

#844

### Task Description

As an AaC Test Developer, I want to expand the test coverage in my unit tests, so that I can have more confidence in the robustness of my code.

### Acceptance Criteria

- [ ] Update test logic as necessary to expand test coverage. 
  - [ ] Happy path and sad path at a minimum
- [ ] (Stretch) Incorporate a behave feature file for testing if applicable.
- [ ] Run `flake8` on the targeted test directory and address findings.
- [ ] Document any identified gaps.

### Unresolved questions

_No response_

bunchw:
# Description

Improve Tests in Context Unit Tests

# Linked Items:

Closes/Fixes/Resolves #853

### Added

- _Describe any new features._

### Changed

### Deprecated

### Removed

### Fixed

- _Describe any bug fixes._

### Security

# Checklist:

- [x] I bumped the version.

### DIFF
--- a/python/tests/test_aac/context/test_definition.py
+++ b/python/tests/test_aac/context/test_definition.py
@@ -1,6 +1,8 @@
 from unittest import TestCase
 
+from aac.context.definition import Definition
 from aac.context.language_context import LanguageContext
+from aac.in_out.files.aac_file import AaCFile
 
 
 class TestDefinition(TestCase):
@@ -18,10 +20,56 @@ class TestDefinition(TestCase):
 
         definition = context.get_definitions_by_name("abstract")
         self.assertEqual(definition[0].get_root_key(), "modifier")
-        
+
     def test_to_yaml(self):
         context = LanguageContext()
         definition = context.get_definitions_by_name("Schema")
         yaml = definition[0].to_yaml()
         self.assertIn("schema:", yaml)
         self.assertIn("name: Schema", yaml)
+
+    def test_definition_fail_incorrect_types(self):
+        context = LanguageContext()
+        schema_definition = context.get_definitions_by_name("Schema")[0]
+        with self.assertRaises(TypeError):
+            # This is probably the worst way to make a definition, it should normally be handled by Language Context
+            definition = Definition(
+                name=123,
+                package=schema_definition.package,
+                content=schema_definition.content,
+                source=schema_definition.source,
+                structure=schema_definition.structure,
+            )
+        with self.assertRaises(TypeError):
+            definition = Definition(
+                name=schema_definition.name,
+                package=123,
+                content=schema_definition.content,
+                source=schema_definition.source,
+                structure=schema_definition.structure,
+            )
+
+VALID_AAC_YAML_CONTENT = """
+schema:
+  name: TestSchema
+  description: |
+    This is a test schema.
+  fields:
+    - name: string_field
+      type: string
+      description: |
+        This is a test field.
+    - name: integer_field
+      type: integer
+      description: |
+        This is a test field.
+    - name: boolean_field
+      type: boolean
+      description: |
+        This is a test field.
+    - name: number_field
+      type: number
+      description: |
+        This is a test field.
+""".strip()
+
--- a/python/tests/test_aac/context/test_definition.py
+++ b/python/tests/test_aac/context/test_definition.py
@@ -20,6 +20,31 @@ class TestDefinition(TestCase):
         definition = context.get_definitions_by_name("abstract")
         self.assertEqual(definition[0].get_root_key(), "modifier")
 
+    def test_is_import(self):
+        context = LanguageContext()
+        definition = context.parse_and_load(IMPORT_DEFINITION)[0]
+        self.assertTrue(definition.is_import())
+        definition = context.get_definitions_by_name("Schema")[0]
+        self.assertFalse(definition.is_import())
+
+    def test_get_python_module_name(self):
+        context = LanguageContext()
+        definition = context.get_definitions_by_name("Schema")[0]
+        module_name = definition.get_python_module_name()
+        self.assertEqual(module_name, "aac.lang")
+
+    def test_get_python_class_name(self):
+        context = LanguageContext()
+        definition = context.parse_and_load(DEFINITION_VALID)[0]
+        class_name = definition.get_python_class_name()
+        self.assertEqual(class_name, "ModelName")
+
+    def test_get_fully_qualified_name(self):
+        context = LanguageContext()
+        definition = context.get_definitions_by_name("Schema")[0]
+        name = definition.get_fully_qualified_name()
+        self.assertEqual(name, "aac.lang.Schema")
+
     def test_to_yaml(self):
         context = LanguageContext()
         definition = context.get_definitions_by_name("Schema")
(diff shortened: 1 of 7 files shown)


python/tests/test_aac/context/test_definition.py
```diff
@@ -18,10 +19,56 @@ def test_get_root_key(self):
...
+        This is a test field.
+    - name: boolean_field
+      type: boolean
+      description: |
+        This is a test field.
+    - name: number_field
+      type: number
+      description: |
+        This is a test field.
+""".strip()
```
lizzcondrey, line 74: Am I missing where this is being used?

python/tests/test_aac/context/test_definition.py
```diff
@@ -18,10 +19,56 @@ def test_get_root_key(self):
...
+                structure=schema_definition.structure,
+            )
+        with self.assertRaises(TypeError):
+            definition = Definition(            # noqa: F841
+                name=schema_definition.name,
+                package=123,
+                content=schema_definition.content,
+                source=schema_definition.source,
+                structure=schema_definition.structure,
+            )
```
lizzcondrey, line 49: Agree with your comment, but good to check that the class constructor does properly handle, even if it is "managed" by a different layer.

Can test the content piece with bad yaml content possibly?

python/tests/test_aac/context/test_definition.py
lizzcondrey, line 1: Be sure to test the other public methods provided by the `Definition` class. Should be good to not have to test the ones with the underscores.

python/tests/test_aac/context/test_definition_parser.py
lizzcondrey, line 1: Again make sure to exercise the other methods as well. We tested the top level with the load definitions, but good to test the major underlying methods as well, any "get" type functions is a good rule of thumb for underlying methods worth testing.

python/tests/test_aac/context/test_language_context.py
```diff
@@ -107,6 +109,53 @@ def test_get_values_by_field_chain(self):
...
+        with self.assertRaises(ParserError):
+            context = LanguageContext()
+            context.parse_and_load("definition")
+
+    def test_get_python_type_from_primitive_no_type(self):
+        context = LanguageContext()
+        with self.assertRaises(LanguageError):
+            context.get_python_type_from_primitive("type")
+
+    # Should this method return an error of some kind when it finds no definitions of a name?
```
lizzcondrey, line 142: Yes it should

python/tests/test_aac/context/test_language_context.py
```diff
@@ -107,6 +109,53 @@ def test_get_values_by_field_chain(self):
...
+        context = LanguageContext()
+        definitions = context.get_definitions_by_name("invalid name")
+        self.assertEqual(definitions, [])
+        with self.assertRaises(TypeError):
+            definitions = context.get_definitions_by_name(123)
+
+    # Should this method return an error of some kind when it finds no definitions of a root?
+    def test_get_definitions_by_root_fail(self):
+        context = LanguageContext()
+        definitions = context.get_definitions_by_root("invalid root")
```
lizzcondrey, line 153: Yes it should

python/tests/test_aac/context/test_language_context.py
lizzcondrey, line 1: Looks like missing tests for:
get_aac_core_definitions
is_aac_instance
remove_definitions

lizzcondrey: And I would clean up the ordering of the tests, have them follow same order as the methods and group pass/fail together

python/tests/test_aac/context/test_source_location.py
```diff
@@ -5,5 +5,9 @@
 class TestSourceLocation(TestCase):
     def test_source_location(self):
         loc = SourceLocation(1, 2, 3, 4)
-        
-        self.assertTrue(isinstance(loc.to_tuple(), tuple))
+
+        self.assertIsInstance(loc.to_tuple(), tuple)
```
lizzcondrey, line 9: I would expand this test to check the mapping of the tuple entries as well. So, the `SourceLocation` class has each entry of the tuple designated as line, column, position, and span. I would do a check for the mapping results.

python/tests/test_aac/context/test_no_undefined_fields.py
lizzcondrey, line 1: I would also add a case of testing when an undefined top definition level field is tried to be used.

lizzcondrey (changes requested):

python/tests/test_aac/context/test_definition_parser.py
bunchw: I would think all of the methods in definition parser get tested in language context right?  Everything in definition parser gets called from a method in language context.

python/tests/test_aac/context/test_no_undefined_fields.py
bunchw: Could you explain this a little more?  Not sure I understand what you mean.

lizzcondrey: Making a definition that has some undefined field besides the ones at the top level like:
```
name:
description:
package:
root:
```

python/tests/test_aac/context/test_definition_parser.py
lizzcondrey: The methods in definition parser do help provide some of the content that is returned with methods in language context, but don't think they all have a direct equivalent. Should review and confirm. If they do, then I think that is fine to only explicitly test the language context methods since our current design is that you would interact with the language context object vs the definition parser.

python/tests/test_aac/context/test_definition.py
```diff
@@ -18,10 +19,88 @@ def test_get_root_key(self):
...
-        
+
+    def test_is_import(self):
+        context = LanguageContext()
+        definition = context.parse_and_load(IMPORT_DEFINITION)[0]
+        self.assertTrue(definition.is_import())
+        definition = context.get_definitions_by_name("Schema")[0]
+        self.assertFalse(definition.is_import())
+
+    def test_get_python_module_name(self):
```
lizzcondrey, line 30: Be sure to do fail cases for these as well. You checked both states for `is_import`

python/tests/test_aac/context/test_source_location.py
lizzcondrey: @bunchw don't forget to check the source location mapping of the tuple entries.

python/tests/test_aac/context/test_no_undefined_fields.py
lizzcondrey: The way you leveraged this definitely works as a valid badly defined fields.

But, meant something more like

```
name: name
    description: description content
    not_real_field: field content
    package: aac.lang
```

python/tests/test_aac/context/test_language_context.py
lizzcondrey: Guessing the question came up because you weren't getting a dedicated error message? If so, capture it as a gap. They should all return an internal LanguageError or ParserError.

bunchw: I was wondering, because I wasn't sure if we wanted the `get_definitions_by_name()` method to act more like a search function that returns an empty list if nothing is found (Which is what it does now), or if we always want it to find something.

python/tests/test_aac/context/test_no_undefined_fields.py
bunchw: Is that not covered by the first half of the test_no_extra_fields test?  The one that uses BAD_EXTRA_FIELDS?

lizzcondrey: I didn't expand the view enough to see that line, yes, it is covered with the first portion of the BAD_EXTRA_FIELDS.

lizzcondrey (changes requested):

lizzcondrey (approved):

PR #899, opened 2024-10-29
No ext for final unit tests

jasoncbryant:
# Description

Updated No Ext for Final Unit Tests

# Linked Items:

Closes/Fixes/Resolves (https://github.com/DevOps-MBSE/AaC/issues/855)

### Changed

- Removed unneeded imports. Split existing test method into two methods, a happy path and a sad path. Added an additional sad path case. Added a feature file.

# Checklist:

- [X] I updated project documentation to reflect my changes.
- [X] My changes generate no new warnings.
- [X] I updated new and existing unit tests to account for my changes.
- [X] I linked the associated item(s) to be closed.
- [X] I bumped the version.
- [X] I added the labels corresponding to my changes.


python/features/plugins/no_ext_for_final.feature
lizzcondrey, line 1: You'll need to modify the structure of your when statements to be able to work with the modified dynamic when steps now incorporated for testing our commands.

Take a look at the other feature files that got added with Will's last PR. 

They got re-established to meet the four command execution cases we had discussed this morning during backlog refinement.

Also, not entirely certain why it looks like pretty much all of the `check.feature` file got duplicated here? I would have expected to really only see the feature file working with the `shapes.aac` file since I believe that is our only example model that leverages the `extension` concept.

python/tests/test_aac/plugins/no_ext_for_final/test_no_ext_for_final.py
```diff
@@ -1,38 +1,55 @@
...
 
+
+        #  Off-nominal (expected fail) tests with a invalid schema definition
+        #  TestSchema defined twice
+        definitions = context.parse_and_load(root_schema_with_bad_ext_duplicate)
+        result = no_extension_for_final(definitions[0].instance, definitions[0], schema_definition.instance)
+        self.assertTrue(result.is_success())
+        result = no_extension_for_final(definitions[1].instance, definitions[1], schema_definition.instance)
+        self.assertFalse(result.is_success())
+        #  context.remove_definitions(definitions) #  Intentionally commented out
```
lizzcondrey, line 50: So is this actually needing to remain? If so, need more explanation for why its staying but commented out. If not, needs to be removed.

python/tests/test_aac/plugins/no_ext_for_final/test_no_ext_for_final.py
```diff
@@ -1,38 +1,55 @@
...
         result = no_extension_for_final(definitions[0].instance, definitions[0], schema_definition.instance)
         self.assertTrue(result.is_success())
         result = no_extension_for_final(definitions[1].instance, definitions[1], schema_definition.instance)
         self.assertTrue(result.is_success())
         context.remove_definitions(definitions)
 
-        definitions = context.parse_and_load(root_schema_with_bad_ext)
+
+    #  Off-nominal (expected fail) tests
+    def test_no_extension_for_final_offnominal(self):
```
lizzcondrey, line 25: I think another valid off nominal case would be trying to extend something that is not defined. So, like you typo the model name you're trying to extend for instance.

python/features/plugins/no_ext_for_final.feature
lizzcondrey: Cleaning this file up appropriately should then allow the test step of the pipeline to properly pass.

lizzcondrey (changes requested):

python/tests/test_aac/plugins/no_ext_for_final/test_no_ext_for_final.py
jasoncbryant: Yes! I had that one once but lost it when I did a commit in gitpod but not the push. Added back in. Good catch

python/tests/test_aac/plugins/no_ext_for_final/test_no_ext_for_final.py
jasoncbryant: Removed.

python/features/plugins/no_ext_for_final.feature
jasoncbryant: I've reduced the feature to only the shapes.aac portion. I'll look at the four command execution cases from home.

python/tests/test_aac/plugins/no_ext_for_final/test_no_ext_for_final.py
lizzcondrey: Looks like the does not exist test had a similar line commented out, so that one now needs removed.

python/features/plugins/no_ext_for_final.feature
lizzcondrey: For this test set, should only have to worry about the case of a command with arguments and with no flags. The flags are evaluated at the `check` feature level since the flag actually belongs to `check`. 

Also, need to make changes to the provided example shape model to do an actual exercise of the extension capability. Right now, it's just a good implementation and the `bad_shapes.aac` file improperly applies a constraint to cause the failure. It's what was used to also exercise the `Mutually Exclusive Field` constraint since it's the only example definition that currently leveraged it.

lizzcondrey (changes requested):

python/features/plugins/no_ext_for_final.feature
jasoncbryant: The commented out line was from the copy-paste for the missing test case. That's been corrected.

jasoncbryant: Leave the example shape files as they are and make new copies for this feature file to call? I'm wary of making changes to the existing files since they are used in numerous places.

lizzcondrey: So, you should be fine to use the good shapes file as is. Can make a copy and alter for the fail case of altering the applied extension field to trigger the constraint to fail and alter the name to be about "bad shape extension" or the like.

lizzcondrey (approved):

PR #931, opened 2024-11-13
Req Parsing to Context

### LINKED ISSUE #841: 🐛 Definitions not being parsed and added to context
### Pre-requisites

- [X] I checked the documentation and found no answer.
- [X] I checked to make sure that this issue has not already been filed.

### What version of the AaC package are you using? Check your version with the command: aac version.

0.4.11

### Description

When I parse a model with req traces I'm not getting a listing of all the defined reqs.  There's nothing different or strange about how these are defined that I can see.  But interestingly it is consistent, run after run the same reqs fail to be recognized.

### Reproduction URL

_No response_

### Reproduction steps

```bash
Parse the sample here:  https://github.com/DevOps-MBSE/aac-doc-mdl/blob/main/tests/samples/trade_study/doc/cookie_trade_study.aac

Get all the `req` items from the context and you'll see there are some missing items.
```

### Screenshots

```bash
If applicable, add screenshots to help explain your problem.
![DESCRIPTION](LINK.png)
```

### Logs

```bash
Please copy and paste any relevant log output. This will be automatically formatted into code, so no need for backticks.
```

### Pertinent Files

Tip: You can attach images or log files by clicking this area to highlight it and then dragging files in.

bunchw:
# Description

Bug fix for requirement definitions being overwritten in the fully qualified name dictionary.

# Linked Items:

Closes/Fixes/Resolves #841 

### Added

- _Describe any new features._

### Changed

### Deprecated

### Removed

### Fixed

- _Describe any bug fixes._

### Security

# Checklist:

- [ ] My changes generate no new warnings.
- [ ] I updated new and existing unit tests to account for my changes.
- [x] I bumped the version.

### DIFF
--- a/python/features/plugins/trade_study_requirements.feature
+++ b/python/features/plugins/trade_study_requirements.feature
@@ -38,8 +38,3 @@ Feature:  Trade study document model with requirements example
             |GCD-012    |
             |GCD-013    |
             |GCD-014    |
-
-    Scenario:  Check the echo model
-        Given I have the "./features/trade_study/doc/cookie_trade_study.aac" file
-        When I run the "check" command with arguments "./features/trade_study/doc/cookie_trade_study.aac" and with no flags
-        Then I should receive a message that the command was successful
--- a/python/features/steps/general_purpose_steps.py
+++ b/python/features/steps/general_purpose_steps.py
@@ -190,7 +190,6 @@ def check_req_id(context, req_id):
         context: Active context to check against.
         req_id: A list of req id values.
     """
-    print("DEBUG: running 'I should have requirement ids' step_impl for {req_id}")
     parsed_ids = []
     for definition in context.aac_loaded_definitions:
         if definition.get_root_key() == "req":


lizzcondrey:
I don't see changes to tests for helping confirm and exercise this altered handling.

python/src/aac/context/definition_parser.py
```diff
@@ -813,11 +813,16 @@ def load_definitions(
...
-            definition.get_fully_qualified_name(): definition
-            for definition in self.parsed_definitions + self.context.get_definitions()
-            if "package" in definition.structure[definition.get_root_key()]
-        }
+        self.fully_qualified_name_to_definition = {}
+        for definition in self.parsed_definitions + self.context.get_definitions():
+            if "package" in definition.structure[definition.get_root_key()]:
+                fully_qualified_name = definition.get_fully_qualified_name()
+
+                # This is so requirements specifically do not get overwritten.  Although other definition types will still get overwritten, so we may need to find a better solution eventually.
```
lizzcondrey, line 821:
```suggestion
                # This is so requirements specifically do not get overwritten.  Although other definition types might still get overwritten, so we may need to find a better solution eventually.
```

python/src/aac/context/definition_parser.py
```diff
@@ -830,8 +835,13 @@ def load_definitions(
             definition.source.is_loaded_in_context = True
             result.append(definition)
             self.context.context_instance.definitions.add(definition)
+            fully_qualified_name = f"{definition.package}.{definition.name}"
+
+            # This is so requirements specifically do not get overwritten.  Although other definition types will still get overwritten, so we may need to find a better solution eventually.
```
lizzcondrey, line 840:
```suggestion
            # This is so requirements specifically do not get overwritten.  Although other definition types might still get overwritten, so we may need to find a better solution eventually.
```

python/src/aac/context/definition_parser.py
lizzcondrey: We don't know for certain that other definitions are being overwritten at this time, the combination of the package reference and name reference was part of mitigating that potential for definitions, but it wasn't accommodating requirement definitions being slightly different.

python/src/aac/context/definition_parser.py
```diff
@@ -813,11 +813,16 @@ def load_definitions(
...
-            for definition in self.parsed_definitions + self.context.get_definitions()
-            if "package" in definition.structure[definition.get_root_key()]
-        }
+        self.fully_qualified_name_to_definition = {}
+        for definition in self.parsed_definitions + self.context.get_definitions():
+            if "package" in definition.structure[definition.get_root_key()]:
+                fully_qualified_name = definition.get_fully_qualified_name()
+
+                # This is so requirements specifically do not get overwritten.  Although other definition types will still get overwritten, so we may need to find a better solution eventually.
+                if definition.get_root_key() == "req":
```
lizzcondrey, line 822: Need to do a similar evaluation with `req_spec` and be sure they aren't going to encounter a similar overwrite issue. May help shape what a "better solution" might need to look like if we confirm it is possible for other definitions to be overwritten. The original bug was around requirements, and both `req` and `req_spec` fall in that bucket.

python/src/aac/context/definition_parser.py
```diff
@@ -813,11 +813,16 @@ def load_definitions(
             for definition in self.parsed_definitions + self.context.get_definitions()
             if definition.get_root_key() == "primitive"
         }
-        self.fully_qualified_name_to_definition = {
-            definition.get_fully_qualified_name(): definition
-            for definition in self.parsed_definitions + self.context.get_definitions()
-            if "package" in definition.structure[definition.get_root_key()]
-        }
+        self.fully_qualified_name_to_definition = {}
+        for definition in self.parsed_definitions + self.context.get_definitions():
```
lizzcondrey, line 817: For consistency, I would also adjust the logic structuring for the primitive type assignment above so that the flow and leverage is better understood/more readable

python/src/aac/context/definition_parser.py
bunchw: They will get overwritten in the fully qualified name dictionary if they share a name.  I believe this only affects datarefs, so we just haven't run into it yet.  Its just that requirements are the most common dataref objects that could share a name.

lizzcondrey: But the logic within the `load_definitions` method when it is assigning the key (our formatting of a name for referencing) for the dictionary is originally intending to use a combination of the `definition.package` and `definition.name` to help mitigate the name collision potential. 

Our requirement data structures do not have a concept of `package` to help ensure unique names, so we went a step further and for `req` definitions are appending the `req_id` portion as well now. 

And a `dataref` is intended to drill to a specific portion of the `definition` for referencing which generally is just a `name`, so we need to look into what version of `name` is being leveraged. Is it the `name` assigned to the `name` field in the definition or is it the `name` key created for the dictionary registry of loaded definitions.

python/src/aac/context/definition_parser.py
bunchw: Im not really sure how I would do that.  req_spec does not have an ID or any other identifying fields other than its name.  Could we have a package field for req specs?

lizzcondrey (changes requested):

lizzcondrey:
Also, there will be some extra branch clean up with this when we are ready to merge. Looks like you had two branches created around the concept, and the example branch JD had made should be deleted as well since the files to exercise the problem got brought over to this branch.

lizzcondrey (approved):

PR #934, opened 2024-12-13
Improvement of LanguageError Lexeme Usage

### LINKED ISSUE #819: 📋 [TASK] - Improvement of LanguageError Lexeme Usage
### Parent Feature

#914 

### Task Description

As an AaC Developer, I want to implement the identified improvements for the LanguageError usage of the Lexemes, so that I provide a reliable and useful tool for end users.

### Acceptance Criteria

- [ ] LanguageError Lexeme usage is modified as necessary to better support utilization. 
  - [ ] get_location_str method is modified to return SourceLocation
  - [ ] DefinitionParser class is updated to use modified get_location method
  - [ ] Accurate location data from an erroneous parsing is provided to ExecutionMessage
- [ ] Unit tests for the Lexemes provided to LanguageError are updated.
- [ ] Documentation about utilizing Lexemes is expanded to cover the LangaugeError Lexeme usage layer. 

### Unresolved questions

_No response_

bunchw:
# Description

Improvement of LanguageError Lexeme Usage

# Linked Items:

Closes/Fixes/Resolves #819 

### Added

init and str methods to language error, and a language error test class

### Changed

### Deprecated

### Removed

### Fixed

- _Describe any bug fixes._

### Security

# Checklist:

- [x] I bumped the version.

### DIFF
--- a/python/src/aac/context/language_error.py
+++ b/python/src/aac/context/language_error.py
@@ -7,7 +7,7 @@ from typing import Optional
 class LanguageError(Exception):
     """A base class representing a language error condition."""
     message: str = attrib(validator=validators.instance_of(str))
-    location: Optional[str] = attrib(validator=validators.optional(validators.instance_of(str)))
+    location: str = attrib(validator=validators.instance_of(str))
 
     def __init__(self, message: str, location: str):
         """
--- a/python/src/aac/context/language_error.py
+++ b/python/src/aac/context/language_error.py
@@ -4,7 +4,7 @@ from attr import attrib, attrs, validators
 
 @attrs(slots=True)
 class LanguageError(Exception):
-    """A base class representing a language error condition."""
+    """A base class representing a language error."""
     message: str = attrib(validator=validators.instance_of(str))
     location: str = attrib(validator=validators.instance_of(str))
 
@@ -13,8 +13,8 @@ class LanguageError(Exception):
         An init constructor for the Language Error Class.
 
         Args:
-            message(str): The error message.
-            location(str): The source location of the error.
+            message (str): The error message.
+            location (str): The source location of the error.
         """
         self.message = message
         self.location = location
--- a/python/src/aac/context/language_error.py
+++ b/python/src/aac/context/language_error.py
@@ -1,6 +1,5 @@
 """An exception class representing a language error condition."""
 from attr import attrib, attrs, validators
-from typing import Optional
 
 
 @attrs(slots=True)
--- a/python/src/aac/context/language_error.py
+++ b/python/src/aac/context/language_error.py
@@ -9,9 +9,22 @@ class LanguageError(Exception):
     message: str = attrib(validator=validators.instance_of(str))
     location: Optional[str] = attrib(validator=validators.optional(validators.instance_of(str)))
 
-    def __init__(self, message, location):
+    def __init__(self, message: str, location: str):
+        """
+        An init constructor for the Language Error Class.
+
+        Args:
+            message(str): The error message.
+            location(str): The source location of the error.
+        """
         self.message = message
         self.location = location
 
-    def __str__(self):
-        return (f"Error: {self.message}. Location: {self.location}")
+    def __str__(self) -> str:
+        """
+        Method for defining how Language Error looks when represented as a string.
+
+        Returns:
+            The Language Error represented as a string.
+        """
+        return f"Error: {self.message}. Location: {self.location}"
--- a/python/src/aac/context/language_error.py
+++ b/python/src/aac/context/language_error.py
@@ -8,3 +8,10 @@ class LanguageError(Exception):
     """A base class representing a language error condition."""
     message: str = attrib(validator=validators.instance_of(str))
     location: Optional[str] = attrib(validator=validators.optional(validators.instance_of(str)))
+
+    def __init__(self, message, location):
+        self.message = message
+        self.location = location
+
+    def __str__(self):
+        return (f"Error: {self.message}. Location: {self.location}")
--- a/python/src/aac/context/lexeme.py
+++ b/python/src/aac/context/lexeme.py
@@ -20,6 +20,19 @@ class Lexeme:
     source: str = attrib(validator=validators.instance_of(str))
     value: str = attrib(validator=validators.instance_of(str))
 
+    def __init__(self, location: SourceLocation, source: str, value: str):
+        """
+        An init constructor for the Language Error Class.
+
+        Args:
+            location (SourceLocation): The position in the source file.
+            source (str): The source file.
+            value (str): The value stored in the lexeme.
+        """
+        self.location = location
+        self.source = source
+        self.value = value
+        
     def __eq__(self, __o) -> bool:
         """Return whether this Lexeme is the same as __o."""
         return (
--- a/python/src/aac/context/lexeme.py
+++ b/python/src/aac/context/lexeme.py
@@ -32,7 +32,7 @@ class Lexeme:
         self.location = location
         self.source = source
         self.value = value
-        
+
     def __eq__(self, __o) -> bool:
         """Return whether this Lexeme is the same as __o."""
         return (
--- a/python/src/aac/context/lexeme.py
+++ b/python/src/aac/context/lexeme.py
@@ -22,7 +22,7 @@ class Lexeme:
 
     def __init__(self, location: SourceLocation, source: str, value: str):
         """
-        An init constructor for the Language Error Class.
+        An init constructor for the Lexeme Class.
 
         Args:
             location (SourceLocation): The position in the source file.
--- a/python/src/aac/execute/command_line.py
+++ b/python/src/aac/execute/command_line.py
@@ -25,6 +25,7 @@ from aac.execute.aac_execution_result import (
 )
 from aac.context.language_error import LanguageError
 from aac.context.language_context import LanguageContext
+from aac.in_out.files.aac_file import AaCFile
 from aac.in_out.parser._parser_error import ParserError
 
 from typing import Callable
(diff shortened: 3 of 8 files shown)


python/src/aac/context/language_error.py
```diff
@@ -8,3 +8,23 @@ class LanguageError(Exception):
...
+        An init constructor for the Language Error Class.
+
+        Args:
+            message(str): The error message.
+            location(str): The source location of the error.
+        """
+        self.message = message
+        self.location = location
+
+    def __str__(self) -> str:
```
lizzcondrey, line 23: While this isn't an incorrect way to set this, if we are wanting it to be a way of fetching a formatted string of the LanguageError content, you would not want to house it in a "protected" method like this with the name mangling. You would have it as a public method without the name mangling (so no surrounding underscores). That way people would understand it is meant to be used and is not just some internal representation. If you are intending for it to only be used by internal logic and not an exposed usable method for end users, then you would make it "private" and only have a preceding underscore to the start of the name, i.e. _get_language_error_as_str. So, we just need to decide how this is wanting to be used and adjust the method name accordingly.

python/src/aac/context/language_error.py
```diff
@@ -8,3 +8,23 @@ class LanguageError(Exception):
...
+    def __init__(self, message: str, location: str):
+        """
+        An init constructor for the Language Error Class.
+
+        Args:
+            message(str): The error message.
+            location(str): The source location of the error.
+        """
+        self.message = message
+        self.location = location
```
lizzcondrey, line 21: Should still provide the type hints here. 
```suggestion
        self.message: str = message
        self.location: Optional[str] = location
```

lizzcondrey: Though, we need to discuss if we really want the location portion to be an optional attribute for the error. That is what allows for us to pass `None` in the `location` entries of a `LanguageError`. Do we really want a `None` or do we want a paired down `location` of at minimum returning the file that triggered the error?

python/tests/test_aac/context/test_language_error.py
```diff
@@ -0,0 +1,16 @@
+from unittest import TestCase
+
+from aac.context.language_context import LanguageContext
+from aac.context.language_error import LanguageError
+
+class TestLanguageError(TestCase):
+    def test_language_error(self):
```
lizzcondrey, line 7: With having the `location` set as an optional, should also test that it handles fine receiving `None`, if we don't decide to move away from that implementation.

lizzcondrey (changes requested):
Make sure to bump version.

python/src/aac/context/language_error.py
bunchw: __str__ is a special python method, like __init__.  It wont be called.  __str__ is basically whats called when you print() something, it just formats the class into a readable format.

python/src/aac/context/language_error.py
bunchw: If we have None as location, we could instead give it a string like "No Location" or something, and remove the optional from the type.

python/src/aac/context/language_error.py
lizzcondrey: Correct. So, this was not added with the intent of being used elsewhere within the code, but just to leverage that functionality?

bunchw: Yes.

python/src/aac/context/language_error.py
bunchw: Also, I dont think it needs the type hints because its not creating the class variable, its just setting them.  It has type hints above in the class body.

lizzcondrey: > If we have None as location, we could instead give it a string like "No Location" or something, and remove the optional from the type.

So, meaning where we are actually raising the errors if no applicable location provide the "No Location" notice instead?

python/src/aac/context/language_error.py
```diff
@@ -8,3 +8,23 @@ class LanguageError(Exception):
...
     message: str = attrib(validator=validators.instance_of(str))
     location: Optional[str] = attrib(validator=validators.optional(validators.instance_of(str)))
+
+    def __init__(self, message: str, location: str):
+        """
+        An init constructor for the Language Error Class.
+
+        Args:
+            message(str): The error message.
+            location(str): The source location of the error.
```
lizzcondrey, line 18:
```suggestion
            message (str): The error message.
            location (str): The source location of the error.
```

python/src/aac/context/language_error.py
```diff
@@ -8,3 +8,23 @@ class LanguageError(Exception):
     """A base class representing a language error condition."""
```
lizzcondrey, line 8:
```suggestion
    """A base class representing a language error."""
```

lizzcondrey (changes requested):

python/src/aac/execute/command_line.py
```diff
@@ -86,7 +87,7 @@ def wrapper(*args, **kwargs):
                 plugin_name,
                 "exception",
                 ExecutionStatus.GENERAL_FAILURE,
-                [ExecutionMessage(e.message, MessageLevel.ERROR, None, e.location)],
+                [ExecutionMessage(e.message, MessageLevel.ERROR, AaCFile(e.location, True, True), None)],
```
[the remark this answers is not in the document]
bunchw: The fourth argument for execution message is a SourceLocation, the column and row, not the file.  e.location is a string containing the name of the file.

python/src/aac/plugins/check/check_aac_impl.py
```diff
@@ -137,8 +137,8 @@ def check_schema_constraint(
                     # if the field is a list, check each item in the list
                     if type(getattr(check_me, field.name)) != list:
                         raise LanguageError(
-                            f"Value of '{field.name}' was expected to be list, but was '{type(getattr(check_me, field.name))}'. File: '{source_definition.source.uri}'",
-                            None
+                            f"Value of '{field.name}' was expected to be list, but was '{type(getattr(check_me, field.name))}'",
+                            f'{source_definition.source.uri}'
```
lizzcondrey, line 141: Wouldn't drop the additional file designator. Always best to be as explicit in returned errors as you can be.

```suggestion
                            f"Encountered in file: {source_definition.source.uri}"
```

python/tests/test_aac/context/test_language_error.py
lizzcondrey: This can be ignored since we are going to try and move away from having it be an optional and going to force some kind of location provision, even if a notice of something like "No location"

python/src/aac/context/lexeme.py
```diff
@@ -20,6 +20,19 @@ class Lexeme:
     source: str = attrib(validator=validators.instance_of(str))
     value: str = attrib(validator=validators.instance_of(str))
 
+    def __init__(self, location: SourceLocation, source: str, value: str):
+        """
+        An init constructor for the Language Error Class.
```
lizzcondrey, line 25:
```suggestion
        An init constructor for the Lexeme Class.
```

python/src/aac/context/lexeme.py
```diff
@@ -20,6 +20,19 @@ class Lexeme:
     source: str = attrib(validator=validators.instance_of(str))
     value: str = attrib(validator=validators.instance_of(str))
 
+    def __init__(self, location: SourceLocation, source: str, value: str):
```
lizzcondrey, line 23: So, was similar behavior to what was being seen with `LanguageError` found to be happening with the `Lexeme` class then?

lizzcondrey (changes requested):

python/src/aac/execute/command_line.py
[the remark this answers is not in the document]
lizzcondrey: Going to pull that info and make sure it is included with the `ExecutionMessage` task as part of the evaluations and potential changes.

python/src/aac/context/lexeme.py
bunchw: I didn't see anywhere that had that issue, but it's never a bad idea to have an __init__ method anyway, and it could avoid a problem down the line.

python/src/aac/plugins/check/check_aac_impl.py
bunchw: Thats what the __str__ method is for.  Its not neccessary to have a designator before the location because the __str__ has one, and when the error is diplayed it calls the __str__ method to display it.  It will show as 
"Error: {message}. Location: {location}"

bunchw: We can change the actual text of the __str__ method to be anything if we want to also.

lizzcondrey: Okay, so that is part of what I was trying to get at when trying to understand why you implemented the formatting method how you did. I'll try to take a look for making sure this prints out like you expected. Because as it's leveraged right now, the `LanguageError` is just caught as Exceptions then the `message` and `location` portions are accessed from the exception. So, your formatting method won't really play a role as it stands.

lizzcondrey: Is that something you are wanting to change with how the `LanguageError` is leveraged by pieces like `ExecutionMessage` to make use of the `__str__` formatting method you added?

bunchw: I don't think anything would have to change about ExecutionMessage, it doesnt directly interact with language error.  In some places, e.message and e.location are passed into an ExecutionMessage, and that will work as it does currently.

lizzcondrey: So, I am only finding LanguageError to be caught then the individual pieces referenced for displaying the content. Because of that, I don't think your formatting method is ever going to be used. Have you seen different behavior somewhere that I am missing?

bunchw: When a language error is not caught and displays in the terminal, the __str__ method is called.

lizzcondrey: Had side bar further discussing and reviewing the format method and its usage by the tool as backup provision when the exception is just raised and not caught for further control. Should be good to move on from this now.

lizzcondrey (approved):

PR #944, opened 2025-01-21
Context/language_context.py Location Returns

### LINKED ISSUE #927: 📋 [TASK] - Context/language_context.py Location Returns
### Parent Feature

#921

### Task Description

As an AaC Developer, I want to leverage the improved Lexeme returns in the unknown location returns, so that I provide more useful information to end users from the error handling.

### Acceptance Criteria

- [ ] Provide a location in current unknown or missing location returns.
- [ ] Expand unit test for language_context.py to test the location returns.

### Unresolved questions

_No response_

bunchw:
# Description

Context/language_context.py Location Returns

# Linked Items:

Closes/Fixes/Resolves #927 

### Added

Expanded error returns and updated unit tests.

### Changed

### Deprecated

### Removed

### Fixed

- _Describe any bug fixes._

### Security

# Checklist:

- [x] I bumped the version.

### DIFF
--- a/python/src/aac/context/language_context.py
+++ b/python/src/aac/context/language_context.py
@@ -141,10 +141,15 @@ class LanguageContext(object):
             LanguageError: When a unique definition for the AaC Enum is not found, an error message detailing the issue is generated.
         """
         definitions = self.get_definitions_by_name(aac_enum_name)
-        if len(definitions) != 1:
+        if len(definitions) > 1:
             raise LanguageError(
                 f"Unable to identify unique definition for '{aac_enum_name}'.  Found {[definition.name for definition in definitions]}",
                 "No file to reference" if not definitions[0].source or definitions[0].source.uri == "<string>" else definitions[0].source.uri
+            )
+        if len(definitions) < 1:
+            raise LanguageError(
+                f"Unable to identify definition for '{aac_enum_name}'",
+                "No file to reference"
             )
         aac_class = self._get_aac_generated_class(aac_enum_name)
         try:
--- a/python/src/aac/context/language_context.py
+++ b/python/src/aac/context/language_context.py
@@ -45,7 +45,7 @@ class LanguageContext(object):
 
     def get_aac_core_as_yaml(self) -> str:
         """
-        Function to return the AaC language as a yaml string.
+        Function to return the AaC language as a YAML string.
 
         Returns:
             The core AaC definitions file as a string in YAML format.
@@ -81,20 +81,29 @@ class LanguageContext(object):
 
         return aac_class
 
-    def is_aac_instance(self, obj: Any, name: str):
-        """Function to determine if an object is an instance of an AaC class."""
+    def is_aac_instance(self, obj: Any, name: str) -> bool:
+        """
+        Function to determine if an object is an instance of an AaC class.
+
+        Args:
+            obj (Any): Object to check.
+        Function to create a Python object representation of an AaC class and its attributes.
+
+        Returns:
+            A boolean which equals true if the object is an instance of the specified class.
+        """
         return isinstance(obj, self._get_aac_generated_class(name))
 
     def create_aac_object(self, aac_type_name: str, attributes: dict) -> Any:
         """
-        Function to create a python instance of an AaC class and attributes.
+        Function to create a Python object representation of an AaC class and its attributes.
 
         Args:
             aac_type_name (str): The name of the AaC class.
             attributes (dict): A dictionary of attributes for the AaC class.
 
         Returns:
-            An instance of the AaC class.
+            A Python object representing the AaC class.
 
         Raises:
             LanguageError: When a unique definition for the AaC type is not found, an error message detailing the issue is generated.
@@ -116,14 +125,14 @@ class LanguageContext(object):
 
     def create_aac_enum(self, aac_enum_name: str, value: str) -> Any:
         """
-        Function to create a python instance of an AaC enum class and value.
+        Function to create a Python enum for an AaC enum class and value.
 
         Args:
             aac_enum_name (str): The name of the AaC Enum class.
             value (str): The value for the enum.
 
         Returns:
-            An instance of the Enum class set to the specified value.
+            A Python enum set to the specified value.
 
         Raises:
             LanguageError: When a unique definition for the AaC Enum is not found, an error message detailing the issue is generated.
@@ -208,7 +217,7 @@ class LanguageContext(object):
             root_key (str): A root key to search for.
 
         Returns:
-            A list of definitions with the given root.
+           A list of definitions with the given root key.
         """
         result = []
         for definition in self.get_definitions():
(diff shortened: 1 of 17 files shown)


python/tests/test_aac/context/test_language_context.py
lizzcondrey, line 1: So, remember that within unit tests we don't make use of a try/except. You can do a with loop around the executing logic using `assertRaises` of the exception we are evaluating like what was in place originally. Not entirely certain why you swapped approaches.

python/src/aac/context/language_context.py
lizzcondrey, line 1: Are we certain that "No file to reference" is accurate for here? Because based on what checks are happening preceding them, it's evaluations within the file(s) are happening, but that whatever is being referenced as part of the definition is not being found. So, while the other areas are making use of a Lexeme to provide a file location, and there would not be a valid Lexeme to reference here, we now know that a definition keeps track of the file it was parsed from. Would we not reference that data for the location information within these LanguageErrors, like you used in the primitives and check impl?

lizzcondrey: And remember, we are trying to leave files better than we found them. There are missing docstrings throughout that need updated.

lizzcondrey (changes requested):

lizzcondrey:
And make sure you fill out the template content. It gets leveraged for change log updates when we put out new releases so be sure to expand on what altered with the PR under the proper section(s), ie added, changed, removed, etc.

python/tests/test_aac/context/test_language_context.py
```diff
@@ -194,24 +194,35 @@ def test_get_definitions_by_root_fail(self):
             definitions = context.get_definitions_by_name(123)
 
     def test_get_defining_schema_for_root_fail(self):
-        with self.assertRaises(LanguageError):
-            context = LanguageContext()
-            definitions = context.get_defining_schema_for_root("Schema")
+        # with self.assertRaises(LanguageError):
+        #     context = LanguageContext()
+        #     definitions = context.get_defining_schema_for_root("Schema")
```
lizzcondrey, line 199: This commented out portion shouldn't need to remain? Think it got missed in your swap back to the with loops.

python/src/aac/context/language_context.py
bunchw: Where are the missing docstrings?

"No file to reference" is accurate here I believe, because the methods (get_defining_schema_for_root, get_python_type_from_primitive, get_definitions_of_type) are not necessarily called in reference to a definition file, they can be called from anywhere with any string.  Anywhere that does call these methods in reference to a specific definition file would need to have its own language error.

lizzcondrey: From what I am seeing none of the methods have complete docstrings. I can't apply specific comments to them since they are not within range of changes. 

They are all missing info on their args and returns. None of them have the explanation of raised errors/exceptions. 

Remember the docstrings are used as part of our API documentation that is generated when Sphinx does its "scrape" of the code base. So, more information we can provide in method documentation, the better we are from a maintenance stance both within the code base and documentation provision. 

> "No file to reference" is accurate here I believe, because the methods (get_defining_schema_for_root, get_python_type_from_primitive, get_definitions_of_type) are not necessarily called in reference to a definition file, they can be called from anywhere with any string. Anywhere that does call these methods in reference to a specific definition file would need to have its own language error.

Great explanation for providing clarification on the identified layers handling the evaluation of the definition file and division of responsibility. I believe that is a good assessment that the lower/internal layers may not always be working from a file explicitly since they can be leveraged elsewhere to build on and that the higher layers using them that are working with a specific definition file can be responsible for returning the location in its own LanguageError. 

If that distinction and division also makes sense to you, @jasoncbryant , believe it would be worth us capturing in a small write up to keep with our development process pieces to help us adhere to that consistently.

lizzcondrey (changes requested):

lizzcondrey (changes requested):

lizzcondrey (changes requested):
Some grammatical and phrasing clean up. But much better. 

And an issue with the unit test when ran locally.

python/src/aac/context/language_context.py
```diff
@@ -66,12 +81,33 @@ def _get_aac_generated_class(self, name: str) -> Type:
...
 
-    def is_aac_instance(self, obj: Any, name: str):
-        """Function to determine if an object is an instance of an AaC class."""
+    def is_aac_instance(self, obj: Any, name: str) -> bool:
+        """
+        Function to determine if an object is an instance of an AaC class.
+
+        Args:
+            obj (Any): Object to check.
+        Function to create a Python object representation of an AaC class and its attributes.
```
lizzcondrey, line 90:
```suggestion
        name (str):  The AaC class name to be used for comparison.
```

python/src/aac/plugins/aac_primitives/aac_primitive_constraints_impl.py
```diff
@@ -330,7 +331,10 @@ def check_typeref(
     package_name = ".".join(parts[:-1])
     type_name = parts[-1]
 
-    definitions_of_type = context.get_definitions_of_type(package_name, type_name)
+    try:
+        definitions_of_type = context.get_definitions_of_type(package_name, type_name)
+    except LanguageError as e:
+        raise LanguageError(e.message, source.uri)
```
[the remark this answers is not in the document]
lizzcondrey: > That, I think that just needs to be our standard implementation as it covers the scenario you had been describing when I asked about the accuracy of "No file to reference". And leaves us out of having to figure out without doubt which layer is getting to a file location to leverage for all use cases.

With this portion of my comment I had expected to see all of the `LanguageError` instances where you tried to leverage a `source` reference at a minimum to be changed. But wanted to revisit the `LanguageError` discussion from earlier in this PR and decide if that should be our standard implementation for providing a location when no Lexeme to leverage, which is the case of the instances in `LanguageContext` that you adjusted to have "No file to reference".

lizzcondrey (changes requested):

python/src/aac/context/language_context.py
```diff
@@ -136,48 +213,103 @@ def get_definitions_by_name(self, name: str) -> list[Definition]:
...
+
+        Raises:
+            LanguageError: When no primitive type is found with the given name, an error message detailing the issue is generated.
+        """
         primitives = self.get_definitions_by_name(primitive_name)
         if len(primitives) != 1:
             raise LanguageError(
                 message=f"Could not find unique primitive type: {primitive_name} - discovered {[primitive.name for primitive in primitives]}",
-                location="None"
+                location="No file to reference"
```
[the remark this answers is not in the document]
bunchw: Had to split it into two, if it does not find anything here it will cause an out of bounds index otherwise.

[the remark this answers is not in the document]
lizzcondrey: Good catch and adjustment. Makes me wonder if the other length checks would have similar issue though and we've just not triggered that situation yet. Thoughts?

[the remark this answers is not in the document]
bunchw: Maybe, I'll look and see if we have anywhere else like that.

lizzcondrey (approved):
Great alteration of making use of the constant to help streamline the leveraging of the definition sources. Good find! Tests are all still going through fine so that default is coming through as desired.

Issue #924, opened 2024-11-06
📋 [TASK] - Plugins/check_aac_impl Location Returns


lizzcondrey:
### Parent Feature

#921

### Task Description

As an AaC Developer, I want to leverage the improved Lexeme returns in the unknown location returns, so that I provide more useful information to end users from the error handling.

### Acceptance Criteria

- [ ] Provide a location in current unknown or missing location returns.
- [ ] Expand unit test for check_aac to test the location returns.

PR #962, opened 2025-01-28
Improve tests for check aac #924

jasoncbryant:
# Description

Improve tests for check aac. Added several new tests.

# Linked Items:

Closes (https://github.com/DevOps-MBSE/AaC/issues/924) 

### Added

Added additional unit tests for check_aac.
Clarified existing test for one of the three LanguageError cases in check_aac.
A new test for improper YAML input.
A new test with missing required field.
A new test with unknown/unexpected primitive.
A new test with an invalid value.
A new test with an undefined field name.

# Checklist:

- [X] I updated project documentation to reflect my changes.
- [X] My changes generate no new warnings.
- [X] I updated new and existing unit tests to account for my changes.
- [X] I linked the associated item(s) to be closed.
- [X] I bumped the version.
- [X] I added the labels corresponding to my changes.


python/tests/test_aac/plugins/check/test_check_aac.py
```diff
@@ -52,16 +56,99 @@ def test_cli_check_verbose(self):
...
             shutil.copy(aac_file_path, temp_aac_file_path)
 
             check_args = [temp_aac_file_path]
+
             exit_code, output_message = self.run_check_cli_command_with_args(check_args)
             self.assertNotEqual(0, exit_code)
             self.assertIn("was expected to be list, but was", output_message)
+
+    # Test input improper YAML
+    # output msg: The AaC file '/tmp/{random-temp-name}}/my_plugin.aac' could not be parsed.
```
lizzcondrey, line 77: Can add an assertion check of finding the provided files for the test.

python/tests/test_aac/plugins/check/test_check_aac.py
```diff
@@ -52,16 +56,99 @@ def test_cli_check_verbose(self):
...
+
+            check_args = [temp_aac_file_path]
+
+            exit_code, output_message = self.run_check_cli_command_with_args(check_args)
+            # Assert for return code 3 (fail)
+            self.assertEqual(3, exit_code, f"Expected to fail but ran successfully with message: {output_message}")
+            self.assertNotIn("My plugin was successful.", output_message)  # only appears when --verbose is passed in.
+
+    # Test input missing required 'when' field primitive
+    # output msg: Missing required field when.
```
lizzcondrey, line 93: Looks like the remaining tests do not come back with the offending file, catalogue them and see if you can maybe trace them back and see what we've missed altering.

lizzcondrey (changes requested):

python/src/aac/plugins/check/check_aac_impl.py
```diff
@@ -69,12 +91,22 @@ def check_primitive_constraint(
...
-        """Runs all the constraints for a given schema."""
+        """Helper method that runs all the constraints for a given schema.
+
+        Args:
+            source_definition (Definition): Source of the check_me field that we are evaluating
+            check_me (Any):                 The field being checked
+            check_against:                  The schema we are comparing the check_me field against
+
+        Returns:
+            (at check level) ExecutionResult: The result of the checks in this helper method
```
lizzcondrey, line 102: This helper method has a LanguageError within, so need a raises entry as well explaining what causes the exception.

```suggestion
        Returns:
            (at check level) ExecutionResult: The result of the checks in this helper method
            
        Raises:
            LanguageError: When a definition for a field type cannot be found.
            LanguageError: When a field value is expected to be a list and is another type.
          
```

python/src/aac/plugins/check/check_aac_impl.py
```diff
@@ -17,7 +17,17 @@
...
+
+    Args:
+        aac_file (str):         The AaC file being processed
+        fail_on_warn (bool):    Flag to fail when warnings are discovered
+        verbose (bool):         Flag for verbose mode. When true add success messages as encountered.
+
+    Returns:
+        ExecutionResult:        Method result containing: plugin_name ("Check AaC"), "check", status, message
+                                including results from lower level helper methods
+    """
```
lizzcondrey, line 30: Have LanguageError for check method as a whole as well that needs captured in a Raises entry.

```suggestion
    Returns:
        ExecutionResult:        Method result containing: plugin_name ("Check AaC"), "check", status, message
                                including results from lower level helper methods
                                
     Raises:
         LanguageError: When a defining schema for the requested root could not be found. 
    """
```

python/tests/test_aac/plugins/check/test_check_aac.py
```diff
@@ -39,10 +41,12 @@ def test_cli_check(self):
 
             exit_code, output_message = self.run_check_cli_command_with_args(check_args)
             self.assertEqual(0, exit_code, f"Expected success but failed with message: {output_message}")  # asserts the command ran successfully
-            self.assertNotIn("My plugin was successful.", output_message) # only appears when --verbose is passed in.
+            self.assertNotIn("My plugin was successful.", output_message)  # only appears when --verbose is passed in.
```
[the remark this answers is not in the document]
lizzcondrey: The name of the file evaluated is provided with the `output_message` so can do something like
`self.assertIn("my_plugin.aac", output_message)`

python/tests/test_aac/plugins/check/test_check_aac.py
lizzcondrey: Similar here, can make a check of the `output_message` for `my_plugin.aac`

python/tests/test_aac/plugins/check/test_check_aac.py
lizzcondrey: Confirm output messages are not getting the provided file names after making the source changes. If they now do, then make the assertion checks like mentioned previously of looking for the `my_plugin.aac` to be in the `output_messages`. 

If the file name is NOT provided, capture that as being the case and any "bread crumb trail" you might have of where is this error message coming from so that we can assess what we might have missed for making sure the location bubbled up because I believe we have now removed all "None" entries to locations for our LanguageErrors which are majority of where these come from. So, either we missed one, or this is coming from a different error handling than we had been working on correcting so need to be sure it is captured to be done with other error handling work.

lizzcondrey (changes requested):

python/tests/test_aac/plugins/check/test_check_aac.py
```diff
@@ -52,16 +56,99 @@ def test_cli_check_verbose(self):
 
             exit_code, output_message = self.run_check_cli_command_with_args(check_args)
             self.assertEqual(0, exit_code, f"Expected success but failed with message: {output_message}")  # asserts the command ran successfully
-            self.assertIn("was successful.", output_message) # only appears when --verbose is passed in.
+            self.assertIn("my_plugin.aac", output_message)
```
lizzcondrey, line 59: Would want to keep the check for "was successful" as that is evaluating for an altered output as part of the verbose flag.

python/tests/test_aac/plugins/check/test_check_aac.py
```diff
@@ -39,10 +41,12 @@ def test_cli_check(self):
 
             exit_code, output_message = self.run_check_cli_command_with_args(check_args)
             self.assertEqual(0, exit_code, f"Expected success but failed with message: {output_message}")  # asserts the command ran successfully
-            self.assertNotIn("My plugin was successful.", output_message) # only appears when --verbose is passed in.
+            self.assertIn("my_plugin.aac", output_message)
```
lizzcondrey, line 44: Similar here, would want to continue with the other check as well because it lets us confirm that the addition or absence of the verbose flag results in proper output messages.

python/tests/test_aac/plugins/check/test_check_aac.py
lizzcondrey: Status on this?

lizzcondrey (changes requested):

python/tests/test_aac/plugins/check/test_check_aac.py
jasoncbryant: I'm following. This is a mal-formed YAML file. That gets detected fairly early on by code not strictly within check.

python/tests/test_aac/plugins/check/test_check_aac.py
lizzcondrey: Missed adding back the `assertIn` for the `My Plugin was successful.` notice for confirming the verbose handling.

python/tests/test_aac/plugins/check/test_check_aac.py
lizzcondrey: Okay, that is what I had thought when I tried to trace these back, and some of these are exercised with language_context tests since that is where some of these errors originate from.

I don't think it is wrong to confirm their bubble up to check, which is really the only place that returns them. Because even if you run an alternate command, ie one of the generates, it runs check first. So, we'll keep them here as well and just take note that the location information is not coming up with it.

lizzcondrey (approved):

